PR #149, opened 2022-05-17
Update colors' names

smyzin:
- changed colors' names
- update files with new names


tailwind.config.js
```diff
@@ -43,24 +43,24 @@ module.exports = {
...
       },
       colors: {
-        'blue-900': '#0d0030',
-        'blue-800': '#1a0259',
-        'blue-700': '#3a00d6',
-        'blue-600': '#4b61d5',
-        'blue-500': '#15a1fc',
-        'blue-400': '#bff5fa',
+        accent: '#15a1fc',
+        'accent-dark': '#0d0030',
```
T-Damer, line 47: I am not sure if the blue is `accent`, it's mostly used on the background. It more feels like `primary`

tailwind.config.js
```diff
@@ -43,24 +43,24 @@ module.exports = {
...
 
-        orange: '#f1962e',
-        yellow: '#fed823',
-        pink: '#ff7bed',
-        green: '#01feb6',
+        primary: '#fed823',
+        secondary: '#ff7bed',
+        tertiary: '#01feb6',
         white: '#efecd6',
+        extra: '#f1962e',
```
T-Damer, line 57: Let's get rid of this color, we don't use it

tailwind.config.js
```diff
@@ -43,24 +43,24 @@ module.exports = {
...
 
-        pink50: 'rgba(255, 123, 237, 0.5)',
-        yellow50: 'rgba(254, 216, 35, 0.4)',
-        blue50: 'rgba(21, 161, 252, 0.5)',
-        white50: 'rgba(239, 236, 214, 0.4)',
-        green50: 'rgba(1, 254, 182, 0.4)',
+        'accent-semi-transparent': 'rgba(21, 161, 252, 0.5)',
```
T-Damer, line 59: Btw, did you know, that you can use `#fed823` + `50` = `#fed82350` to make it half transparent. Maybe we can use hexes here, it will be better

T-Damer (approved):
Let's have a little renaming and we're gtg

tailwind.config.js
smyzin: Sure, but it's in the design, so I decided to leave it. But anyway, you're right

tailwind.config.js
T-Damer: If we can't, it's fine

tailwind.config.js
smyzin: Changed them

tailwind.config.js
smyzin: Let's leave it for now

PR #160, opened 2022-05-17
Removed default export functions names

T-Damer:
- Removed names from the functions, that have `export default function`
- Changed the `const Component: FC ...` to `export default function Component ...`


src/components/ZKProof.tsx
```diff
@@ -122,4 +122,4 @@ const ZKProof: FC<{ proof?: Proof; contractAddress: string }> = ({
   )
 }
 
-export default ZKProof
+export default ZkProof
```
backmeupplz: The file is called `ZKProop.tsx`

src/components/ZKProof.tsx
```diff
@@ -103,7 +103,7 @@ function useProofContent(
   }
 }
 
-const ZKProof: FC<{ proof?: Proof; contractAddress: string }> = ({
+const ZkProof: FC<{ proof?: Proof; contractAddress: string }> = ({
```
backmeupplz: Should we get rid of FC here?

backmeupplz (changes requested):

PR #170, opened 2022-05-18
Check env variable if something is missing throw an error

T-Damer:
The app won't load if something is missing

### DIFF
--- a/src/env.d.ts
+++ b/src/env.d.ts
@@ -1,0 +1,13 @@
+/// <reference types="vite/client" />
+
+interface ImportMetaEnv {
+  readonly VITE_ENCRYPT_KEY: string
+  readonly VITE_ETH_NETWORK: string
+  readonly VITE_FORTMATIC_KEY: string
+  readonly VITE_BITSKI_CLIENT_ID: string
+  readonly VITE_APP_NAME: string
+  readonly VITE_SC_LEDGER_CONTRACT_ADDRESS: string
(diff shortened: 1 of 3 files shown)


src/env.d.ts
```diff
@@ -0,0 +1,13 @@
+/// <reference types="vite/client" />
```
T-Damer, line 1: Read it from the docs: https://vitejs.dev/guide/env-and-mode.html#intellisense-for-typescript

src/helpers/env.ts
```diff
@@ -1,12 +1,19 @@
...
+  VITE_BITSKI_CLIENT_ID: import.meta.env.VITE_BITSKI_CLIENT_ID,
   VITE_SC_LEDGER_CONTRACT_ADDRESS: import.meta.env
-    .VITE_SC_LEDGER_CONTRACT_ADDRESS as string,
-  VITE_VERIFY_URL: import.meta.env.VITE_VERIFY_URL as string,
+    .VITE_SC_LEDGER_CONTRACT_ADDRESS,
+  VITE_VERIFY_URL: import.meta.env.VITE_VERIFY_URL,
 }
+
+Object.values(env).forEach((value: string, index) => {
+  if (value === '' || value === undefined || value === null)
```
T-Damer, line 15: Actually checking for `''` and `undeinfed` is enough, should I remove `null`?

backmeupplz (approved):

PR #173, opened 2022-05-18
simplify BadgeBlock

upacyxou:
- attempt to simplify badgeblock


backmeupplz (changes requested):
Need to tidy up the classnames order

PR #176, opened 2022-05-18
Refactored card separator

T-Damer:
- Reduced number of available colours
- Shortened props


src/components/CardSeparator.tsx
```diff
@@ -10,7 +11,27 @@ import classnames, {
...
+  number: number
+  gradient: Gradient
+  vertical?: boolean
+  customHeight?: THeight
+}
+
+function getTailwindClassnames(gradient: Gradient) {
+  switch (gradient) {
+    case 'secondary-to-transparent':
+      return 'from-secondary'
```
T-Damer, line 28: I found out, that tailwind applies transparent if we don't specify `to` or `from`, this is why I return only one string here

src/components/CardSeparator.tsx
```diff
@@ -10,7 +11,27 @@ import classnames, {
...
+    case 'accent-to-secondary':
+      return { 'from-accent': true, 'to-secondary': true }
```
T-Damer, line 30: And this is a way to use two tailwind properties, we can't use an array unfortunately, only `{ prop: true, prop2: true}`. We did it previously in some places, it lookes like:
```TS
gradient(color === 'green' ? { 'from-blue': true, 'to-green': true } : undefined)
```

src/components/CardSeparator.tsx
```diff
@@ -10,7 +11,27 @@ import classnames, {
   width,
 } from 'classnames/tailwind'
 
-type Color = 'accent' | 'primary' | 'secondary' | 'tertiary' | 'transparent'
+type Gradient =
+  | 'secondary-to-transparent'
+  | 'accent-to-secondary'
+  | 'accent-to-transparent'
+interface CardSeparatorProps {
```
backmeupplz, line 18: Don't put interfaces at the top of the file, if you only use them once — put them where you use them

src/components/CardSeparator.tsx
```diff
@@ -10,7 +11,27 @@ import classnames, {
...
+  number: number
```
backmeupplz, line 19: What's `number`? Make the name of the parameter less ambiguous

src/components/CardSeparator.tsx
backmeupplz, line 18: Don't forget about the empty lines between declarations

src/components/CardSeparator.tsx
```diff
@@ -10,7 +11,27 @@ import classnames, {
...
+  gradient: Gradient
+  vertical?: boolean
+  customHeight?: THeight
+}
+
+function getTailwindClassnames(gradient: Gradient) {
```
backmeupplz, line 25: Use `classnames` here or whatever function needed from `tailwind-classnames`

src/components/CardSeparator.tsx
```diff
@@ -21,60 +42,19 @@ const connectiveBlock = (vertical = true) =>
...
+export default function (props: CardSeparatorProps) {
   return (
-    <div className={connectiveBlock(vertical)}>
-      {[...Array(number).keys()].map((_, index) => (
-        <div
-          key={index}
-          className={connectiveLine(from, to, vertical, customHeight)}
-        ></div>
+    <div className={connectiveBlock(props.vertical)}>
+      {[...Array(props.number).keys()].map((_, index) => (
```
backmeupplz, line 56: `Array(props.number).fill(null).map(...)`

src/components/CardSeparator.tsx
backmeupplz: Also, name of this function implies that you're getting all the tailwind classnames, you should specify which classnames you're getting here

backmeupplz (changes requested):

src/components/CardSeparator.tsx
```diff
@@ -10,7 +11,23 @@ import classnames, {
...
 } from 'classnames/tailwind'
 
-type Color = 'accent' | 'primary' | 'secondary' | 'tertiary' | 'transparent'
+type Gradient =
+  | 'secondary-to-transparent'
+  | 'accent-to-secondary'
+  | 'accent-to-transparent'
+
+const getTailwindGradient = (gradient: Gradient) =>
+  classnames(
```
backmeupplz, line 20: If it's just one type of classname, you can drop the `classnames`

src/components/CardSeparator.tsx
```diff
@@ -21,61 +38,29 @@ const connectiveBlock = (vertical = true) =>
...
-  from?: Color,
-  to?: Color,
-  vertical = true,
-  customHeight?: number
-) =>
+interface CardSeparatorProps {
+  numberOfLines: number
+  gradient: Gradient
+  vertical?: boolean
+  customHeight?: THeight
```
backmeupplz, line 45: Are you still using this prop?

backmeupplz (changes requested):

src/helpers/bigintConvert.ts
```diff
@@ -98,7 +98,7 @@ export function bnToHex(stringNum: string) {
   return '0x' + hex
 }
 
-function bitnot(bigIntNum: BigInt) {
+function bitnot(bigIntNum: bigint) {
```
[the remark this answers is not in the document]
T-Damer: Yes, GH actions can't pass the lint phase

backmeupplz (approved):

PR #185, opened 2022-05-19
Remove `@types/react`

backmeupplz:
Switch to only using `preact` and `preact/compat' built-in types


backmeupplz:
Pending https://github.com/preactjs/preact/issues/3545

backmeupplz:
Pending https://github.com/preactjs/preact/pull/3546

PR #188, opened 2022-05-20
Figma design check

ohpaul:
- checking designs from figma with frontend


src/components/Badges.tsx
```diff
@@ -45,12 +45,12 @@ function Badges() {
           {shouldNotify ? (
             <BadgesHintCard
               text={
-                <>
+                <div>
```
ohpaul, line 48: `space` isn't recognized for the components unless it's `<div>`, maybe there's a better method

backmeupplz (approved):

backmeupplz (approved):

PR #200, opened 2022-05-23
Removed SimpleBar and used self-written scrollbar

T-Damer:
- Removed plain scrollbar with css
- Added custom scrollbar with css
- Used `onScroll` to detect scroll

Sizes:
image


src/components/Scrollbar.tsx
```diff
@@ -20,15 +17,59 @@ export default function ({
...
+    const scroll = wrapCurrent.scrollTop / numberOfViews
+
+    thumbRef.current.style.top = scroll + 'px'
+  }
+
+  useEffect(() => {
+    const numberOfViews =
+      wrapRef.current.scrollHeight / wrapRef.current.clientHeight
+
+    setThumbHeight(100 / numberOfViews)
```
T-Damer, line 45: This is not responsive, if we'll add some child nodes it won't update the `scrollThumb` height, also seems to be buggy with card, that is more than wrapper

backmeupplz (changes requested):
Let's fix the visual bugs first before merging

backmeupplz (approved):

PR #204, opened 2022-05-24
Fix badge main on small screen

smyzin:
- added a new breakpoint
- added custom styles
- fixed badge name


src/hooks/useBreakpoints.ts
```diff
@@ -20,6 +20,7 @@ export default function () {
   }, [])
 
   return {
+    badgeNameSize: width > smSize && width < 400,
```
T-Damer, line 23: The `useBreakpoint` is used to have different screens breakpoints, maybe you can make a different name for this size?

src/index.css
```diff
@@ -121,6 +121,16 @@ body {
   }
 }
 
+.badgeBlockName {
+  display: -webkit-box;
+  -webkit-line-clamp: 2;
+  -webkit-box-orient: vertical;
```
T-Damer, line 127: The docs say, that you can use `flex-direction`: https://developer.mozilla.org/en-US/docs/Web/CSS/box-orient

src/index.css
```diff
@@ -121,6 +121,16 @@ body {
   }
 }
 
+.badgeBlockName {
+  display: -webkit-box;
+  -webkit-line-clamp: 2;
```
T-Damer, line 126: What about `Ff`? The `webkit` won't work with it

T-Damer: We can use `tailwind` plugin for this: https://tailwindcss.com/docs/plugins#line-clamp

src/index.css
```diff
@@ -121,6 +121,16 @@ body {
   }
 }
 
+.badgeBlockName {
+  display: -webkit-box;
```
T-Damer, line 125: If we can switch to `flex` model, this would be great. If we can't — then let's use this class 👍

T-Damer (changes requested):
Gave comments

src/index.css
smyzin: Set plugin to project

src/index.css
```diff
@@ -121,6 +121,16 @@ body {
...
+  -webkit-line-clamp: 2;
+  -webkit-box-orient: vertical;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  max-width: 6.25rem;
+  height: 2.75rem;
+}
```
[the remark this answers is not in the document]
smyzin: Made it via tailwind

src/index.css
smyzin: Set plugin instead

T-Damer (approved):
Great work :godmode:

PR #206, opened 2022-05-24
Hide scrollbar on Firefox

T-Damer:


src/index.css
```diff
@@ -121,15 +121,9 @@ body {
   }
 }
 
-/* Removes plain scrollbar from Ff*/
 .scrollable-wrapper {
-  scrollbar-width: 0;
   overflow: auto;
```
T-Damer, line 125: Overflow is necessary for self-written custom scrollbar

src/components/Scrollbar.tsx
```diff
@@ -49,7 +48,7 @@ export default function ({
   })
 
   return (
-    <div className={position('relative')}>
+    <div className="relative scrollbar-hide">
```
backmeupplz, line 51: Use `tailwind-classnames`, at least for `relative`

backmeupplz (changes requested):

src/components/Scrollbar.tsx
```diff
@@ -52,7 +52,7 @@ export default function ({
     <div className={position('relative')}>
       <div
         ref={wrapRef}
-        class="scrollable-wrapper"
+        className={overflow('overflow-auto') + ' scrollbar-hide'}
```
T-Damer, line 55: I'd rather use `class` + `className`, but `className` overrides `class`

PR #213, opened 2022-06-08
fix Scrollbar

MixailE:


backmeupplz:
Seems like you're only using `Measure` for `Scrollbar`, add it to the `Scrollbar` component then. Also, what are you measiring? Name your components less ambiguous. Code is more often read than written.

backmeupplz:
Also, don't forget about the PR description, don't be vague

src/components/Measure.tsx
```diff
@@ -0,0 +1,27 @@
+import { JSX } from 'preact'
+import { useLayoutEffect, useRef, useState } from 'react'
+
+export default function Measure({
```
T-Damer, line 4: Don't forget to remove the default exported function name

T-Damer: Let's name it `MeasureContainerHeight` or something like this

src/components/Measure.tsx
```diff
@@ -0,0 +1,27 @@
...
+}: {
+  defaultHeight?: number
+  className?: string
+  children: (args: { height: number }) => JSX.Element
+}) {
+  const container = useRef<HTMLDivElement>(null)
+  const [height, setHeight] = useState(defaultHeight)
+
+  useLayoutEffect(() => {
+    if (container.current) {
```
T-Damer, line 17: Let's remove curly braces after `if`

src/components/Measure.tsx
T-Damer, line 4: Remove the default function name and rename the file, please

src/components/Measure.tsx
T-Damer, line 17: Let's remove curly braces after if

src/components/Scrollbar.tsx
```diff
@@ -13,18 +25,34 @@ interface ScrollbarProps {
...
+const scrollContainer = classnames(
+  display('flex'),
+  flexDirection('flex-col'),
+  flexGrow('grow'),
+  position('relative')
+)
+
 export default function ({
   children,
   maxHeight = 350,
```
T-Damer, line 37: Rename this prop to `initHeight` or something like this, it doesn't represent the actual `maxHeight`

T-Damer (changes requested):
A few more

package.json
```diff
@@ -37,6 +37,7 @@
     "fortmatic": "^2.4.0",
     "preact": "^10.7.3",
     "qr-code-styling": "^1.6.0-rc.1",
+    "react-resize-detector": "^7.1.1",
```
upacyxou, line 40: how much does it weigh in brotli format?

src/components/Scrollbar.tsx
```diff
@@ -1,32 +1,48 @@
...
   const thumbRef = useRef() as MutableRef<HTMLDivElement>
   const [thumbHeight, setThumbHeight] = useState(100)
 
   const { overflows, scrollMaxHeight, isOnTop, isOnBottom } = useIsOverflow(
     wrapRef,
-    maxHeight
+    height
   )
 
+  console.log('height', height, 'scrollMaxHeight', scrollMaxHeight)
```
upacyxou, line 44: do we need this console.log?

src/components/Scrollbar.tsx
```diff
@@ -45,12 +61,13 @@ export default function ({
     const numberOfViews = node.scrollHeight / node.clientHeight
     setThumbHeight(100 / numberOfViews)
   }, [])
+
```
upacyxou, line 64: I guess it could be removed

T-Damer: Doesn't the code above already listens for resizes?

MixailE: It listens for changes inside the scroll

package.json
MixailE: image

T-Damer (approved):
lgtm

PR #221, opened 2022-06-14
Fix text for all minted badges

smyzin:
- updated text on the card when all badges which may be minted are minted
- removed unused `proofsCompleted`
- added `derivativeTokensOwned` variable


T-Damer (approved):
Lgtm

backmeupplz:
Better re-create this PR for `new-verifier` branch

PR #225, opened 2022-06-15
Remove proof on error

smyzin:
- add try/catch
- fix error on init app


ohpaul (approved):
looks good in catching errors during the mint process

smyzin:
Unnecessary for now

PR #226, opened 2022-06-15
Contracts listener

MixailE:
https://trello.com/c/eu8fUX1F/743-sc-add-listeners-to-contractsstore-to-keep-it-updated


src/stores/OriginalContractsStore.ts
```diff
@@ -1,24 +1,61 @@
...
+      OriginalContractsStore.locked = true
+      OriginalContractsStore.contractsOwnedState = await getOwnedERC721(
+        OriginalContractsStore.contractsOwnedStateAccount,
+        OriginalContractsStore.lastBlockId
+          ? OriginalContractsStore.lastBlockId + 1
+          : 0,
+        currentBlockId,
+        currentState
+      )
+      OriginalContractsStore.locked = false
```
backmeupplz, line 44: Should this happen last?

src/stores/OriginalContractsStore.ts
```diff
@@ -1,24 +1,61 @@
...
+import defaultProvider from 'helpers/defaultProvider'
 import getOwnedERC721 from 'helpers/getOwnedERC721'
 
 interface ContractsStoreType {
-  contractsOwned: Promise<string[]>
+  contractsOwnedState: { [token: string]: number }
+  contractsOwnedStateAccount?: string
+  contractsOwned: string[]
+  lastBlockId?: number
+  locked: boolean
```
backmeupplz, line 12: Can we move the util state (e.g. `locked`) to a separate global variable? Don't want it to be a `proxy`

src/stores/OriginalContractsStore.ts
```diff
@@ -1,24 +1,61 @@
...
+      OriginalContractsStore.contractsOwnedState = {}
+      OriginalContractsStore.contractsOwned = []
+    }
+
+    if (!OriginalContractsStore.contractsOwnedStateAccount) return
+
+    if (!OriginalContractsStore.locked) {
+      const currentBlockId = await defaultProvider.getBlockNumber()
+      const currentState = await OriginalContractsStore.contractsOwnedState
+      OriginalContractsStore.locked = true
```
backmeupplz, line 35: Should this happen first?

src/stores/OriginalContractsStore.ts
backmeupplz: What if something throws an exception? Should we do `try/catch/finally`?

src/stores/OriginalContractsStore.ts
```diff
@@ -1,24 +1,61 @@
...
 const OriginalContractsStore = proxy<ContractsStoreType>({
-  contractsOwned: Promise.resolve([]),
-  fetchContractsOwned() {
-    OriginalContractsStore.contractsOwned = WalletStore.account
-      ? getOwnedERC721(WalletStore.account)
-      : Promise.resolve([])
+  locked: false,
+  contractsOwnedState: {},
+  contractsOwned: [],
+  async fetchContractsOwned() {
```
backmeupplz, line 20: Maybe `fetchMoreContractsOwned`?

src/stores/OriginalContractsStore.ts
```diff
@@ -1,24 +1,61 @@
...
+    if (
+      OriginalContractsStore.contractsOwnedStateAccount !== WalletStore.account
```
backmeupplz, line 22: Should this happen in the `account` listener?

backmeupplz (changes requested):

src/stores/OriginalContractsStore.ts
MixailE: Maybe I misunderstood, but checking the selected account should include both listeners

src/stores/OriginalContractsStore.ts
```diff
@@ -1,24 +1,41 @@
...
+    ].getOwnedERC721()
   },
 })
 
 subscribeKey(WalletStore, 'account', () => {
-  OriginalContractsStore.fetchContractsOwned()
+  OriginalContractsStore.fetchMoreContractsOwned()
+})
+
+defaultProvider.on('block', () => {
```
T-Damer, line 37: Do we need to do `() => {...}` here? Maybe we can get rid of `{}`

T-Damer (approved):
lgtm

src/stores/OriginalContractsStore.ts
```diff
@@ -1,24 +1,41 @@
...
+
+    OriginalContractsStore.contractsOwned = await connectedAccounts[
+      WalletStore.account
+    ].getOwnedERC721()
   },
 })
 
 subscribeKey(WalletStore, 'account', () => {
-  OriginalContractsStore.fetchContractsOwned()
+  OriginalContractsStore.fetchMoreContractsOwned()
```
backmeupplz, line 34: Same here, let's do `subscribeKey(WalletStore, 'account', OriginalContractsStore.fetchMoreContractsOwned)`

backmeupplz (changes requested):

backmeupplz (approved):

PR #233, opened 2022-06-17
Temporarily removed Fortmatic and fixed walletconnect

T-Damer:
- Fortmatic doesn't support `goerli` right now
image
- Changed `chainId` for `walletConnect`. Goerli is `5`

### DIFF
--- a/src/helpers/web3Modal.ts
+++ b/src/helpers/web3Modal.ts
@@ -1,12 +1,12 @@
 import CoinbaseWalletSDK from '@coinbase/wallet-sdk'
-import Fortmatic from 'fortmatic'
+// import Fortmatic from 'fortmatic'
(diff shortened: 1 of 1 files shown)


src/helpers/web3Modal.ts
```diff
@@ -1,31 +1,31 @@
...
   network,
   providerOptions: {
-    fortmatic: {
-      package: Fortmatic,
-      options: {
-        key: env.VITE_FORTMATIC_KEY as string,
-        network,
-      },
-    },
+    // fortmatic: {
```
T-Damer, line 17: When we'll use the `mainnet`, we'll need to uncomment it

T-Damer: Actually, `fortmatic` uses `email/number + pwd` to login, should we use it? Feels not `anonymous`

PR #243, opened 2022-06-20
Add extra check on parent height

smyzin:
- add parent height
- add extra calculation


T-Damer (approved):
lgtm, looks pretty hard to understand actually

backmeupplz (approved):

PR #247, opened 2022-06-21
Replace magic numbers and fix behavior

smyzin:
- removed static numbers
- replaced them with real height

### DIFF
--- a/src/components/Scrollbar.tsx
+++ b/src/components/Scrollbar.tsx
@@ -37,10 +37,11 @@ export default function ({
   titlePadding = 0,
 }: ChildrenProp & ScrollbarProps) {
   const { height = 0, ref } = useResizeDetector({ handleWidth: false })
-  const maximumHeightSize = 370
+  const maximumHeightSize = 370 // the maximum height of the block by design
+  const fullHeightInPercentage = 100 // the thumb height in percentages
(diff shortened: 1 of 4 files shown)


src/components/Scrollbar.tsx
```diff
@@ -32,15 +33,16 @@ export default function ({
...
+  titlePadding = 0,
 }: ChildrenProp & ScrollbarProps) {
   const { height = 0, ref } = useResizeDetector({ handleWidth: false })
 
   const thumbRef = useRef() as MutableRef<HTMLDivElement>
   const [thumbHeight, setThumbHeight] = useState(100)
-  const extaHeightParent = parentHeight > 370 ? 40 : 0
-  const extaHeightTitle = extraPadding ? 32 : 0
-  const extraReservedSpace = extaHeightParent + extaHeightTitle
+  const extaBottomPadding = parentHeight > 370 ? bottomPadding : 0
```
upacyxou, line 43: What is `370`? Looks like a magic number

smyzin: It's a maximum height that the scrollbar block may have. Because of static card height, we have to limit the maximum height by design

backmeupplz (approved):

PR #250, opened 2022-06-22
add scwp announce

upacyxou:
- scwp announce at the top of the page added

### DIFF
--- a/src/components/SCWPAnnounce.tsx
+++ b/src/components/SCWPAnnounce.tsx
@@ -1,0 +1,60 @@
+import { AccentText } from 'components/Text'
+import { useSnapshot } from 'valtio'
+import Button from 'components/proofs/Button'
+import Cross from 'icons/Cross'
+import classnames, {
+  alignItems,
+  backgroundColor,
+  display,
+  flex,
+  fontSize,
+  fontWeight,
+  justifyContent,
+  lineHeight,
+  margin,
+  padding,
+} from 'classnames/tailwind'
+import scwpStore from 'stores/SCWPStore'
+import walletStore from 'stores/WalletStore'
+
+export default function () {
+  const { announceClosed } = useSnapshot(scwpStore)
+  const { account } = useSnapshot(walletStore)
+
+  if (announceClosed || account) return null
+
+  return (
+    <div
+      className={classnames(
+        backgroundColor('bg-primary-dimmed'),
+        padding('px-6', 'py-4'),
+        display('flex'),
+        alignItems('items-center'),
+        justifyContent('justify-center'),
+        fontWeight('font-bold'),
+        fontSize('text-sm'),
+        lineHeight('leading-5')
+      )}
+    >
+      <div className={classnames(flex('flex-1'))} />
+      <AccentText small bold color="text-formal-accent">
+        Now introducing zk proof for your work email! Connect wallet to get
+        started.
+      </AccentText>
+      <div
+        className={classnames(
+          display('flex'),
+          flex('flex-1'),
+          margin('ml-auto')
+        )}
+      >
+        <Button
+          onClick={() => (scwpStore.announceClosed = true)}
+          className={classnames(margin('lg:ml-auto', 'ml-6'))}
+        >
+          <Cross />
+        </Button>
+      </div>
+    </div>
+  )
+}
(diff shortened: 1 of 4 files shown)


src/icons/Cross.tsx
```diff
@@ -0,0 +1,30 @@
+export default function () {
+  return (
+    <svg
+      width="18"
+      height="17"
+      viewBox="0 0 18 17"
```
T-Damer, line 6: Can you make it `0 0 18 18`. It should be rectangle (you may try to export it again from figma, if needed)

src/icons/Cross.tsx
```diff
@@ -0,0 +1,30 @@
...
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+    >
+      <line
+        x1="2.41421"
+        y1="1"
+        x2="16.0588"
+        y2="14.6446"
+        stroke="#EFECD6"
```
T-Damer, line 15: Extract this and `stroke-width` into `const strokeStyle = classnames(...)`

src/icons/Cross.tsx
```diff
@@ -0,0 +1,30 @@
...
+        stroke-width="2"
+        stroke-linecap="round"
+      />
+      <line
+        x1="1.94116"
+        y1="14.6446"
+        x2="15.5858"
+        y2="1.00001"
+        stroke="#EFECD6"
```
T-Damer, line 24: Same here with stroke and stroke width

src/components/SCWPAnnounce.tsx
```diff
@@ -0,0 +1,62 @@
...
+  lineHeight('leading-5')
+)
+
+const crossWrapper = classnames(
+  display('flex'),
+  flex('flex-1'),
+  margin('ml-auto')
+)
+
+const cross = classnames(margin('lg:ml-auto', 'ml-6'))
```
T-Damer, line 37: Remove `classnames()` only `margin()`

src/components/SCWPAnnounce.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+export default function () {
+  const { announceClosed } = useSnapshot(scwpStore)
+  const { account } = useSnapshot(walletStore)
+
+  if (announceClosed || account) return null
```
T-Damer, line 43: I guess it's better to return `<></>` at the bottom

src/components/SCWPAnnounce.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  return (
```
T-Damer, line 45: Check for `announceClosed || account` here and return `<></>`. Also, why do we check for account? User with account should see this notification too, I guess

src/icons/Cross.tsx
```diff
@@ -0,0 +1,31 @@
+import classnames, { stroke, strokeWidth, width } from 'classnames/tailwind'
+
+const strokeStle = classnames(
+  strokeWidth('stroke-2'),
+  stroke('stroke-formal-accent')
+)
+
+export default function () {
+  return (
+    <div className={classnames(width('w-6'))}>
```
T-Damer, line 10: Get rid of `classnames` and extract it into `const svgWrapper`, please :)

src/stores/SCWPStore.ts
```diff
@@ -0,0 +1,10 @@
+import { proxy } from 'valtio'
+import PersistableStore from 'stores/persistence/PersistableStore'
+
+class SCWPStore extends PersistableStore {
+  announceClosed = false
+}
+
+const scwpStore = proxy(new SCWPStore()).makePersistent(true)
```
T-Damer, line 8: I guess, that can avoid `vonst` here, simply `export proxy(...).makePersistent(true)`

T-Damer (changes requested):
A little changes and were good to go :)

src/icons/Cross.tsx
```diff
@@ -0,0 +1,33 @@
+import classnames, { stroke, strokeWidth, width } from 'classnames/tailwind'
+
+const strokeStle = classnames(
+  strokeWidth('stroke-2'),
+  stroke('stroke-formal-accent')
+)
+
+const svgWrapper = classnames(width('w-6'))
```
T-Damer, line 8: Let's remove the `classnames` here, only `width()`

T-Damer (approved):
one more and let's go

PR #253, opened 2022-06-22
Zk proof based on email

T-Damer:


src/models/WorkProof.ts
```diff
@@ -0,0 +1,7 @@
+import ProofResult from 'models/ProofResult'
+
+export default interface WorkProof {
+  id: string
```
backmeupplz, line 4: We probably don't have an id

backmeupplz (changes requested):

backmeupplz (approved):

PR #260, opened 2022-06-23
Fixed most suspense errors + add contract name check

smyzin:
- added check for contracts without names
- fixed proofs suspense
- fixed other suspense error

> The problem with `suspense error` still occurs once in the blue moon, and I cannot find the issue but I localized it - it's in the `Badges` component. I guess, we need to change the layout a little

### DIFF
--- a/src/components/proofs/Proof.tsx
+++ b/src/components/proofs/Proof.tsx
@@ -1,6 +1,5 @@
 import { AccentText, BodyText } from 'components/Text'
 import { JSX } from 'preact'
-import { useSnapshot } from 'valtio'
 import { useState } from 'react'
 import Button from 'components/proofs/Button'
 import Complete from 'icons/Complete'
@@ -54,7 +53,6 @@ function useProofContent(
   color: 'text-accent' | 'text-secondary' | 'text-tertiary'
   content: JSX.Element | null
 } {
-  const { account } = useSnapshot(WalletStore)
   const [isGenerating, setIsGenerating] = useState(false)
 
   if (isGenerating)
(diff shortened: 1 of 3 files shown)


T-Damer (approved):
👏

src/stores/ContractNamesStore.ts
```diff
@@ -16,7 +17,9 @@ const ContractNamesStore = proxy<ContractNamesStoreType>({
       return
     }
     const contract = ERC721__factory.connect(address, defaultProvider)
-    ContractNamesStore.contractNames[address] = contract.name()
+    ContractNamesStore.contractNames[address] = contract.callStatic
+      .name()
+      .catch(() => truncateMiddleIfNeeded(address, 17))
```
backmeupplz, line 22: Why 17? Also, this is a `ContractNamesStore`, not a `ContractNamesOrAddressesStore`. Return `undefined`, instead of the address if `name()` doesn't exist

src/components/proofs/Proof.tsx
```diff
@@ -54,7 +53,6 @@ function useProofContent(
   color: 'text-accent' | 'text-secondary' | 'text-tertiary'
   content: JSX.Element | null
 } {
-  const { account } = useSnapshot(WalletStore)
```
backmeupplz, line 57: This change seems harmful. We should be using snapshots while rendering, not proxy states.

backmeupplz (changes requested):

src/components/proofs/Proof.tsx
smyzin: Changed it

smyzin: If I return this line back, it will throw an error on switching accounts. This line causes that effect for the list of proofs

backmeupplz: You need to find a different solution, you **should not** use proxy state inside components, only the snapshots. The fact that snapshots break the UI tells us that there is an underlying problem that needs to be solved, not patched like this.

PR #261, opened 2022-06-23
make ContractNamesStore persistable

MixailE:


src/stores/ContractNamesStore.ts
```diff
@@ -1,23 +1,39 @@
 import { ERC721__factory } from '@big-whale-labs/seal-cred-ledger-contract'
 import { proxy } from 'valtio'
+import PersistableStore from 'stores/persistence/PersistableStore'
 import defaultProvider from 'helpers/defaultProvider'
 
-interface ContractNamesStoreType {
-  contractNames: {
+class ContractNamesStore extends PersistableStore {
+  saveContractNames = {} as {
```
backmeupplz, line 7: Probably should be called `savedContractNames`

backmeupplz (changes requested):

PR #264, opened 2022-06-24
Input state and tiny message

smyzin:
- created `TinyMessage` component
- stylized input component
- added new text component


src/components/Text.tsx
```diff
@@ -243,3 +243,23 @@ const proofSectionTitle = classnames(
...
+    textColor(
+      color === 'error'
+        ? 'text-error'
+        : color === 'primary'
+        ? 'text-primary-semi-dimmed'
+        : 'text-formal-accent'
+    ),
+    fontFamily('font-primary'),
+    fontSize('text-xs'),
+    lineHeight('!leading-3')
```
T-Damer, line 258: Why do we need `!` here?

src/components/TextForm.tsx
```diff
@@ -20,13 +23,16 @@ export default function ({
     <>
       <Input
         type="text"
+        isError={!!error}
```
T-Damer, line 26: You're using `!!error` twice, let's extract it into the `const hasError`

src/icons/Exclamation.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+const svgElementClasses = classnames(
+  stroke('stroke-formal-accent'),
+  strokeWidth('stroke-1')
+)
+
+export default function () {
+  return (
+    <svg
+      width="13"
```
T-Damer, line 11: Let's remove `width` and `height` from here and wrap the `svg` into `div` with width and height defined

src/icons/Exclamation.tsx
```diff
@@ -0,0 +1,45 @@
...
+      height="13"
+      viewBox="0 0 13 13"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+    >
+      <circle
+        cx="6.50004"
```
T-Damer, line 18: We can round those numbers I think, try doing `cx="6.5" cy="6.5"` etc

T-Damer (changes requested):
mostly fine, let's change a few things

src/components/Text.tsx
smyzin: Agree, removed it

src/components/TextForm.tsx
smyzin: Done

src/icons/Exclamation.tsx
smyzin: Rounded them. Thanks

src/icons/Exclamation.tsx
smyzin: I don't think it's necessary to do it right now because we would rescale it as I see it according to the design. If it's essential, I may come up few ideas

src/components/proofs/EmailProof.tsx
```diff
@@ -67,7 +69,16 @@ export default function () {
     }
   }
 
-  async function onGenerateProof(secret?: string) {
+  async function onGenerateProof(secret: string) {
+    setError(undefined)
```
upacyxou, line 73: Do we need to call this on every `onGenerateProof` call?

1Damer (approved):
lgtm, only one small thing left :)

backmeupplz (approved):

src/components/Text.tsx
```diff
@@ -243,3 +243,23 @@ const proofSectionTitle = classnames(
...
+        ? 'text-error'
+        : color === 'primary'
+        ? 'text-primary-semi-dimmed'
+        : 'text-formal-accent'
+    ),
+    fontFamily('font-primary'),
+    fontSize('text-xs'),
+    lineHeight('leading-3')
+  )
+export function TinyText({
```
upacyxou, line 260: In my opinion, `tinyText` is not self-explanatory, why can't we just make a tiny version of any other texts? Maybe it's more like a `hintText`?

T-Damer (approved):
lgtm

src/components/proofs/EmailProof.tsx
```diff
@@ -67,7 +69,16 @@ export default function () {
...
+
+    if (!checkDomainToken(secret)) {
```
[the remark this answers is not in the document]
smyzin: Changed it, thx

src/components/proofs/EmailProof.tsx
smyzin: Yes,  we need to reset that state. Also, moved it lower near `setLoading` state to reset it if  everything is okay

src/components/Text.tsx
T-Damer: I do agree with @smyzin, it's better to try to create unified `TinyText`, than many different small texts

upacyxou (approved):

PR #266, opened 2022-06-24
Fix long email domains

smyzin:
- fixed question mark component
- polished left block with emails
- fixed right badge with long email domain


1Damer (approved):
lgtm

PR #268, opened 2022-06-24
Email public page

T-Damer:
- Added reversedEmailLedger into SCStore
- Checked if we have `email` contract now and displayed proper text
- Fixed smile styles

### DIFF
--- a/src/components/owned/OwnerInfo.tsx
+++ b/src/components/owned/OwnerInfo.tsx
@@ -8,6 +8,7 @@ import SealCredStore from 'stores/SealCredStore'
 import Smile from 'icons/Smile'
 import Title from 'components/Title'
 import classnames, {
+  alignItems,
   borderColor,
   display,
   flexDirection,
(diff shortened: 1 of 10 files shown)


src/components/owned/OwnerInfo.tsx
```diff
@@ -58,16 +63,18 @@ export default function ({
...
       <BodyText>
-        This is a zkNFT derivative. It means this person has been verified to
-        own at least one ‘
+        {emailRecord
+          ? 'This is a zkNFT derivative of a work email. It means this person has been verified to work at '
+          : 'This is a zkNFT derivative. It means this person has been verified to own at least one '}
+        ‘
         <ExternalLink
-          url={getEtherscanAddressUrl(record.originalContract.address)}
+          url={getEtherscanAddressUrl(record.derivativeContract.address)}
```
MixailE, line 71: There should be the original address, if it is not there, then you can display the domain

backmeupplz (approved):

PR #269, opened 2022-06-24
Badge icons for EmailProofs + polished old icon + added one more random ERC721 icon

T-Damer:


src/components/badges/BadgeBlock.tsx
```diff
@@ -53,7 +56,7 @@ function Badge({ proof }: { proof: BaseProof }) {
 
   return (
     <BadgeCard
-      top={<BadgeIcon />}
+      top={isEmailProof ? <EmailBadge /> : <BadgeIcon />}
```
backmeupplz, line 59: Let's rename the `BadgeIcon` as well?

src/components/badges/BadgeBlock.tsx
```diff
@@ -64,12 +67,12 @@ function Badge({ proof }: { proof: BaseProof }) {
...
             {proof instanceof ERC721Proof ? (
               <ContractName address={proof.contract} />
-            ) : proof instanceof EmailProof ? (
+            ) : isEmailProof ? (
               proof.domain
             ) : (
               'Unknown'
             )}{' '}
-            (derivative)
+            {isEmailProof ? '(email)' : '(derivative)'}
```
backmeupplz, line 75: I think keeping `derivative` is better?

T-Damer: Inside the design, we have `(email)`
image

src/components/badges/BadgeBlock.tsx
```diff
@@ -64,12 +67,12 @@ function Badge({ proof }: { proof: BaseProof }) {
           <>
             {proof instanceof ERC721Proof ? (
```
smyzin, line 68: Maybe it's better to set `proof instanceof ERC721Proof` in the `const` variable. As you did with `isEmailProof`

PR #270, opened 2022-06-24
added link to announcement banner

tuansydau:
## What
* Added a link to /app in the announcement banner
* Made the banner disappear after the link is pressed

### DIFF
--- a/src/components/Announcement.tsx
+++ b/src/components/Announcement.tsx
@@ -15,6 +15,7 @@ import classnames, {
   lineHeight,
   margin,
   padding,
+  textDecoration,
   transitionProperty,
   translate,
 } from 'classnames/tailwind'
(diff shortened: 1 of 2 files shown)


src/components/Announcement.tsx
```diff
@@ -53,6 +56,17 @@ export default function () {
       <div className={classnames(flex('flex-1'))} />
       <AccentText small bold color="text-formal-accent">
         {announcementText}
+        <a
+          href="/app"
+          className={announcementLink}
+          onClick={() => {
+            setTimeout(() => {
+              announcementStore.announcementClosed = true
+            }, 75)
```
backmeupplz, line 65: Is it absolutely necessary to use the timeout here?

backmeupplz (approved):

src/components/Announcement.tsx
tuansydau: No timeout needed, good catch

PR #274, opened 2022-06-25
add BadgeTitle

MixailE:


src/components/badges/BadgeTitle.tsx
```diff
@@ -0,0 +1,36 @@
...
+      <ExternalLink url={getEtherscanAddressUrl(derivativeAddress)}>
+        <ContractName address={derivativeAddress} />
+      </ExternalLink>
+    )
+
+  if (!proof) return null
+
+  return (
+    <>
+      <ProofName badge={proof} /> (derivative)
```
T-Damer, line 33: We also have `(email)` text in designs
image

T-Damer: Also, what will this function return if you pass `derivativeAddress`, but won't pass `proof`?

T-Damer (approved):
only small fixes needed

src/components/badges/BadgeTitle.tsx
MixailE: For simplicity added just return of the name of the derivative contract

PR #276, opened 2022-06-26
fix Announcement

MixailE:


src/components/Announcement.tsx
```diff
@@ -41,31 +40,23 @@ const crossWrapper = classnames(
...
 )
 
-const announcementLink = classnames(textDecoration('underline'))
-
 export default function () {
   const { announcementClosed, announcementText } =
     useSnapshot(announcementStore)
-  const { account } = useSnapshot(walletStore)
   const [animate, setAnimate] = useState(false)
+  const location = useLocation()
```
T-Damer, line 47: Maybe we should put it inside `<App />`, not check the location?

src/components/Announcement.tsx
```diff
@@ -41,31 +40,23 @@ const crossWrapper = classnames(
...
 
-  const closedOrAccountConnected = announcementClosed || account
+  const announcementPage = '/app'
+  if (announcementClosed || location.pathname === announcementPage) return null
 
-  return closedOrAccountConnected ? null : (
+  return (
     <div id="bottom-bar" className={announceWrapper(animate)}>
       <div className={classnames(flex('flex-1'))} />
```
T-Damer, line 54: We can remove `classnames` from here

src/components/Text.tsx
```diff
@@ -193,8 +193,10 @@ export function LinkText({
   gradientFrom,
   gradientTo,
   children,
+  blank,
```
T-Damer, line 196: Let's call it `targetBlank`, it's not obvious what is `blank`

src/components/Text.tsx
```diff
@@ -206,8 +208,10 @@ export function LinkText({
       className={linkText(bold, color, gradientFrom, gradientTo)}
       href={url}
       title={title}
-      rel="noopener noreferrer"
-      target="_blank"
+      {...(blank && {
```
T-Damer, line 211: Let's remove `rel` attribute, because [all modern browsers include it automatically for `_blank`](https://stackoverflow.com/a/50709724/14385332) + let's rewrite this line:
```TS
target={targetBlank && '_blank'}
```
or if it won't work:
```TS
target={targetBlank ? '_blank' : '_self'}
```

src/components/landing/LearnMoreCard.tsx
```diff
@@ -13,7 +13,7 @@ export default function () {
       <PrivacyDeath />
       <div className={bigWhaleLabsText}>
         <BodyText>
-          <LinkText color="text-primary" url={`https://bigwhalelabs.com`}>
+          <LinkText blank color="text-primary" url={`https://bigwhalelabs.com`}>
```
T-Damer, line 16: I think we should pass `url="https://bigwhalelabs.com"`, not `url={`https://bigwhalelabs.com`}`

T-Damer (approved):
lgtm

PR #296, opened 2022-06-29
Fix position and wrap only question

smyzin:
- added listeners to show/hide the tooltip
- set the tooltip to `?` symbol


backmeupplz (approved):

T-Damer (approved):
noice

PR #306, opened 2022-06-30
add BadgeSection

MixailE:


src/components/ContractName.tsx
```diff
@@ -2,19 +2,35 @@ import { Suspense, memo } from 'react'
...
+  truncate,
+  clearType,
+}: ContractNameProps) {
+  const { emailDerivativeContracts = [], ERC721derivativeContracts = [] } =
+    useSnapshot(SealCredStore)
   const { contractNames } = useSnapshot(ContractNamesStore)
-  const contractName = contractNames[address]
-  if (!contractNames[address]) ContractNamesStore.fetchContractName(address)
+  let contractName = contractNames[address]
+  if (!contractName) ContractNamesStore.fetchContractName(address)
```
T-Damer, line 25: What if there's no `contractName`, then it tries to fetch it, no `contractName` again, you'll get an infinite loop here?

src/components/badges/BadgeSection.tsx
```diff
@@ -0,0 +1,28 @@
...
+import ChildrenProp from 'models/ChildrenProp'
+import classnames, {
+  display,
+  gap,
+  gridAutoRows,
+  gridTemplateColumns,
+} from 'classnames/tailwind'
+
+const badgesList = classnames(
+  display('grid'),
```
T-Damer, line 12: We use same styles for in `List.tsx`, maybe it's better to reuse them?

T-Damer (changes requested):
Mostly fine, let's check if we can make it better :)

src/components/ContractName.tsx
MixailE: No, the `contractName` will be requested only once)

src/components/badges/BadgeSection.tsx
MixailE: They are no longer in `List.tsx` 🤔

T-Damer (approved):
lgtm

smyzin (approved):
Looks good

PR #307, opened 2022-06-30
Make css scrollbar instead of self written

T-Damer:
- Remove self written
- This won't work on FF

### DIFF
--- a/src/components/Scrollbar.tsx
+++ b/src/components/Scrollbar.tsx
@@ -26,7 +26,8 @@ const scrollContainer = classnames(
   display('flex'),
   flexDirection('flex-col'),
   flexGrow('grow'),
-  position('relative')
+  position('relative'),
+  overflow('overflow-y-auto')
 )
 
 export default function ({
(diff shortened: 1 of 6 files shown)


src/index.css
```diff
@@ -4,6 +4,8 @@
 
 html {
   background-color: #0d0030;
+  scrollbar-color: #fed823 #8ea5b5;
```
T-Damer, line 7: This actually works in FF

src/index.css
```diff
@@ -4,6 +4,8 @@
 
 html {
   background-color: #0d0030;
+  scrollbar-color: #fed823 transparent;
```
T-Damer, line 7: This actually works in FF

src/index.css
```diff
@@ -33,6 +35,44 @@ body {
   }
 }
 
+::-webkit-scrollbar {
+  width: 25px;
```
T-Damer, line 39: Just like in design

src/index.css
```diff
@@ -33,6 +35,44 @@ body {
...
+    rgba(0, 0, 0, 0) 47%,
+    #fed823 47%,
+    #fed823 51%,
+    rgba(0, 0, 0, 0) 52%
+  );
+}
+::-webkit-scrollbar-thumb {
+  background: #fed823;
+  border-radius: 20px;
+  background-clip: content-box;
```
T-Damer, line 54: A trick to reduce the size of `thumb`

src/index.css
```diff
@@ -33,6 +35,44 @@ body {
...
+  border: 8px solid transparent;
+  box-shadow: inset #fed823;
+}
+::-webkit-scrollbar-thumb:hover {
+  background: #fed950;
+  border-radius: 20px;
+  background-clip: content-box;
+  border: 7.5px solid transparent;
```
T-Damer, line 62: I think this looks fine

https://user-images.githubusercontent.com/49658988/176898450-69e6c1b6-e04f-4c15-8752-dbbc3af43043.mp4

e

src/index.css
```diff
@@ -33,6 +35,44 @@ body {
...
+}
+
+.scroll-shadow-box:before {
```
T-Damer, line 65: Rewrite styles of original library to make the shadow more noticable

src/components/Card.tsx
```diff
@@ -65,13 +69,22 @@ const cardColor = (color?: Color) => {
   )
 }
 
+const appStyles = classnames(
+  space('space-y-6'),
+  display('flex'),
+  flexDirection('flex-col'),
+  alignItems('items-stretch'),
+  minHeight('sm:min-h-card', 'min-h-fit')
```
T-Damer, line 77: Basically this applies the box to grow between `minHeight` and `maxHeight`

T-Damer: + we don't need to dance with `height` and `maxHeight`, only min-max. Max is defined without `forApp` prop

src/components/Fade.tsx
```diff
@@ -15,9 +15,9 @@ const fade = (bottom = true) =>
       bottom ? 'from-transparent' : 'from-primary-dark',
       bottom ? 'to-primary-dark' : 'to-transparent'
     ),
-    position('absolute'),
-    inset(bottom ? 'bottom-0' : 'top-0', 'left-0', 'right-0'),
-    height('h-3'),
+    position('sticky'),
```
T-Damer, line 18: Supercool, it solves all pain created by JS

src/components/Scrollbar.tsx
```diff
@@ -1,115 +1,38 @@
...
-        {children}
-        {isOnBottom && (fade === 'both' || fade === 'bottom') && (
-          <Fade bottom />
-        )}
-      </div>
+    <div
+      ref={wrapperRef}
+      className={wrapperStyles}
+      style={{
+        marginRight: isOverflow && hasWebKit ? '-0.7rem' : '0rem',
```
T-Damer, line 30: Basically, we needed the `maxHeight` only for `<Fade />`, because it was `absolute`, aligned to `outerDiv` (not this one). This `innerDiv` didn't know about the `<Card />` wrapper component and tried to overflow using `outerDiv`, but it wasn't configured to overflow properly (I am not sure, if we can make `parent` overflow if the children overflows 😄)

src/components/Scrollbar.tsx
```diff
@@ -1,115 +1,38 @@
...
-  }, [
-    titlePadding,
-    parentHeight,
-    bottomPadding,
-    ref,
-    wrapperRef,
-    wrapperRef.current?.scrollHeight,
-    wrapperRef.current?.clientHeight,
-  ])
+  const hasWebKit = useHasWebkit()
```
T-Damer, line 22: We don't need `marginRight` for FF

src/components/badges/List.tsx
```diff
@@ -55,31 +45,31 @@ function BadgeListSuspended() {
...
-      />
-      <BadgeSection
-        title="Email derivatives"
-        minted={ownedEmailDerivativeContracts}
-        proofs={proofsAvailableToMint.filter(
-          (proof) => proof instanceof EmailProof
-        )}
-      />
-    </div>
+    <Scrollbar>
```
T-Damer, line 48: We shouldn't use `Scrollbar` outside the really scrollable area, it will cause problems, like the `<HintCard />` above was scrollable

PR #312, opened 2022-06-30
Quick fix to preventing loading state on signing

smyzin:
- removed `walletLoading` from `signMessage` method


MixailE (approved):

src/stores/WalletStore.ts
```diff
@@ -53,16 +53,9 @@ class WalletStore extends PersistableStore {
   }
 
   async signMessage(message: string) {
-    if (!provider) throw new Error('No provider')
```
backmeupplz, line 56: Don't remove this, throw an error

PR #314, opened 2022-07-01
Fix truncation for badge name

smyzin:
- added additional styles for truncation


src/components/badges/BadgesOwnedForContractLoading.tsx
```diff
@@ -11,7 +11,6 @@ export default function ({ contractAddress }: { contractAddress: string }) {
     <BadgeWrapper minted={false}>
       <BadgeCard
         top={<Erc721Badge />}
-        leanLeft
```
T-Damer, line 14: Maybe we don't need this prop in `BadgeCard` at all?

T-Damer (approved):
lgtm, hope it works fine on small devices :)

PR #319, opened 2022-07-04
Fix lines for Safari

T-Damer:


src/icons/SuperOrbWithConnectors.tsx
```diff
@@ -140,15 +120,13 @@ const SuperOrb = () => (
       d="M73.5,47.78"
       transform="translate(-48.02 -4.86)"
       fill="none"
-      stroke-linecap="round"
-      stroke="url(#a)"
+      className={strokeTertiary}
```
T-Damer, line 123: We can't use `svg className={stroke...}` here, because it will outline the orb gradient themselves

MixailE (approved):

PR #325, opened 2022-07-05
Fix button glow got clipped by overflow-hidden

T-Damer:


smyzin (approved):
Tested in Safari, Chrome, Ff. Seems to work fine ;)

PR #331, opened 2022-07-06
Fixes email flow

smyzin:
- fixed tiny message
- fixed buttons


src/components/Button.tsx
```diff
@@ -55,6 +58,11 @@ const commonClasses = (
     boxShadow('shadow-2xl', 'hover:shadow-lg', 'active:shadow-md'),
     width(fullWidth ? 'w-full' : 'w-fit'),
     textAlign(center ? 'text-center' : undefined),
+    fontSize(small ? 'text-sm' : 'text-lg'),
+    lineHeight(small ? 'leading-5' : 'leading-7'),
+    type !== 'tertiary'
+      ? padding(small ? 'py-2' : 'py-4', small ? 'px-4' : 'px-6')
```
T-Damer, line 64: I think doing the `small ? {'py-2': true, 'px-4': true} : ...` is faster than checking for `small` two times

src/components/Text.tsx
```diff
@@ -286,14 +286,19 @@ const tinyText = (color: 'base' | 'primary' | 'error', primary?: boolean) =>
...
+  fontPrimary,
   children,
-}: { color?: 'base' | 'primary' | 'error'; primary?: boolean } & ChildrenProp) {
-  return <div className={tinyText(color || 'base', primary)}>{children}</div>
+}: {
+  color?: 'base' | 'primary' | 'error'
+  fontPrimary?: boolean
+} & ChildrenProp) {
+  return (
+    <div className={tinyText(color || 'base', fontPrimary)}>{children}</div>
```
T-Damer, line 302: Why don't we do `color = 'base'` in the props?

T-Damer (approved):
Mostly cool, let's make sure it's clean

src/components/Button.tsx
smyzin: Idk about that. Thanks)

src/components/Text.tsx
smyzin: You are right. Changed it

PR #333, opened 2022-07-06
Apply transaction for ContractSynchronizer

MixailE:


T-Damer (approved):
lgtm

src/helpers/EmailBadgeBuilder.ts
```diff
@@ -14,7 +14,6 @@ export default class EmailBadgeBuilder extends BaseBadgeBuilder {
   }
 
   create(proof: EmailProof) {
-    if (!proof.result) return
```
smyzin, line 17: Why did you remove this check?

src/helpers/ERC721BadgeBuilder.ts
```diff
@@ -14,7 +14,6 @@ export default class ERC721BadgeBuilder extends BaseBadgeBuilder {
   }
 
   create(proof: ERC721Proof) {
-    if (!proof.result) return
```
smyzin, line 17: Why did you remove this check?

src/helpers/ContractSynchronizer.ts
```diff
@@ -1,15 +1,37 @@
...
+  isTransferEvent,
+  parseLogData,
+} from 'helpers/getOwnedERC721'
 
 export default class ContractSynchronizer {
   account: string
   locked = false
   synchronizedBlockId?: number
   addressToTokenIds: { [address: string]: Set<string> } = {}
+  skipTransactions = new Set<string>()
```
smyzin, line 12: I don't think we will have a variety of hashes. Do we really need `Set`? If so, okay then

smyzin (commented):
Interesting solution

smyzin (approved):

src/helpers/ERC721BadgeBuilder.ts
MixailE: I throw an error in mint function

src/helpers/ContractSynchronizer.ts
MixailE: It would be possible to replace this with an array, I'll think about it

PR #336, opened 2022-07-07
Fix Link font size and add truncate if no ens

smyzin:
- made link font size smaller on tiny screens
- add truncate if no ens address on large screens


T-Damer (approved):
lgtm

PR #345, opened 2022-07-08
Added selection colors and special case

T-Damer:


src/components/Text.tsx
```diff
@@ -61,7 +61,11 @@ const accentText = (
   shadow?: TDropShadow
 ) =>
   classnames(
-    textColor(color),
+    textColor(
+      color === 'text-primary-semi-dimmed'
+        ? { 'selection:text-primary': true, 'text-primary-semi-dimmed': true }
```
T-Damer, line 66: Maybe we should extract this object into `const primaryWithSelection` and share it across the project from `Text.ts`?

MixailE (approved):

PR #349, opened 2022-07-08
Fixed button with arrow and didn't break loading spacing

T-Damer:


src/components/Button.tsx
```diff
@@ -176,13 +176,13 @@ export default function ({
     >
       {loading && <Loading small={small} />}
       {showContent && (
-        <div className={space('space-x-2')}>
+        <>
```
T-Damer, line 179: We already have `space-x-2` inside the `<button>` styles, but it didn't work because...

src/components/Button.tsx
```diff
@@ -176,13 +176,13 @@ export default function ({
...
           {typeof children === 'string' && gradientFont ? (
             <span className={textGradient({ type, loading, disabled })}>
               {children}
             </span>
           ) : (
-            children
+            <div>{children}</div>
```
T-Damer, line 185: ...because this was used in DOM as a plain text, not box element

smyzin (approved):
lgtm

PR #351, opened 2022-07-08
Fixed sticky hover on mobiles, button will turn into green only on few seconds

T-Damer:
- Explained inside the code

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -47,6 +47,12 @@ body {
       0 0/2rem 2rem;
   }
 }
+@media (hover: hover) and (pointer: fine) {
+  .hover-tertiary:hover {
+    color: #01feb6;
+    text-decoration: underline;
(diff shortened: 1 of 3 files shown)


src/index.css
```diff
@@ -47,6 +47,12 @@ body {
       0 0/2rem 2rem;
   }
 }
+@media (hover: hover) and (pointer: fine) {
```
T-Damer, line 50: This will check, if the browser supports real hover (mobile only emulate it) and will apply `:hover` for desktops

T-Damer: And no, it's impossible to do `lg:hover:text-tertialry`, `active:text-tertiary`, we can use `useBreakpoints` for that, but I think this is more native way, because some desktops do support touchscreens (which will emulate `:hover` too)

PR #357, opened 2022-07-09
Persist contract store

MixailE:


src/helpers/ContractSynchronizer.ts
```diff
@@ -5,15 +5,64 @@ import getOwnedERC721, {
...
+    account: string,
+    addressToTokenIds: { [address: string]: Set<string> } = {},
+    synchronizedBlockId?: number
+  ) {
     this.account = account
+    this.synchronizedBlockId = synchronizedBlockId
+    this.addressToTokenIds = addressToTokenIds
+  }
+
+  static fromJSON({
```
T-Damer, line 31: Let's name it `contractSynchronizerFromJson`

src/helpers/ContractSynchronizer.ts
```diff
@@ -5,15 +5,64 @@ import getOwnedERC721, {
...
+      {}
+    )
+    return new ContractSynchronizer(
+      account,
+      addressToTokenSetIds,
+      synchronizedBlockId
+    )
+  }
+
+  toJSON() {
```
T-Damer, line 54: And this is `contractSynchronizerToJson`

src/stores/ContractsStore.ts
```diff
@@ -1,64 +1,73 @@
...
-    if (!ContractsStore.connectedAccounts[WalletStore.account])
-      ContractsStore.connectedAccounts[WalletStore.account] =
-        new ContractSynchronizer(WalletStore.account)
-    if (
-      accountChange ||
-      !ContractsStore.connectedAccounts[WalletStore.account] ||
-      !ContractsStore.contractsOwned
-    ) {
-      ContractsStore.contractsOwned = ContractsStore.connectedAccounts[
+    if (!this.connectedAccounts[WalletStore.account]) {
```
T-Damer, line 45: You can do assignment without `{}`

src/stores/ContractsStore.ts
```diff
@@ -1,64 +1,73 @@
...
+}
+
+const contractsStore = proxyWithComputed(new ContractsStore(), {
+  contractsOwned: (state) =>
+    walletStore.account && state.connectedAccounts[walletStore.account]
+      ? state.connectedAccounts[walletStore.account].contractsOwned
+      : [],
+}).makePersistent(true)
 
 subscribeKey(WalletStore, 'account', () =>
```
T-Damer, line 64: Maybe we can pass `subscribeKey(WalletStore, 'account', contractStore.fetchMoreContractsOwned)`?

src/stores/ContractsStore.ts
```diff
@@ -1,64 +1,73 @@
...
-  contractsOwned: Promise.resolve([]),
-  currentBlock: undefined,
+  replacer = (key: string, value: unknown) => {
+    const disallowList = [] as string[] // ['contractsOwned']
+    return disallowList.includes(key) ? undefined : value
+  }
+
+  reviver = (key: string, value: unknown) => {
+    if (key === 'connectedAccounts') {
+      return Object.entries(
```
T-Damer, line 21: Why do we need to convert the object into `array-like` object?

src/helpers/ContractSynchronizer.ts
```diff
@@ -5,15 +5,64 @@ import getOwnedERC721, {
...
+    return {
+      account: this.account,
+      synchronizedBlockId: this.synchronizedBlockId,
+      addressToTokenIds: Object.entries(this.addressToTokenIds).reduce(
```
T-Damer, line 58: It's pretty hard to understand on what is going on here, maybe you can simplify it?

src/helpers/ContractSynchronizer.ts
```diff
@@ -5,15 +5,64 @@ import getOwnedERC721, {
...
+  static fromJSON({
+    account,
+    synchronizedBlockId,
+    addressToTokenIds,
+  }: {
+    account: string
+    synchronizedBlockId: number
+    addressToTokenIds: { [address: string]: string[] }
+  }) {
+    const addressToTokenSetIds = Object.entries(addressToTokenIds).reduce(
```
T-Damer, line 40: It's pretty hard to understand on what is going on here, maybe you can simplify it? Maybe we can extract it into a function?

T-Damer (changes requested):
looks good, let's simplify it :)

src/helpers/ContractSynchronizer.ts
MixailE: I see no reason to add this before the method name, this is a similar method as `toString`, `toNumber`, `toObject`

src/stores/ContractsStore.ts
MixailE: Here we can’t do this, the context is lost

src/helpers/ContractSynchronizer.ts
MixailE: I see no reason to add this before the method name, this is a similar method as `toString`, `toNumber`, `toObject`

src/helpers/ContractSynchronizer.ts
MixailE: Added `transformObjectValues` ​​function :)

src/helpers/ContractSynchronizer.ts
MixailE: Added `transformObjectValues` ​​function :)

T-Damer (approved):

backmeupplz (approved):

PR #358, opened 2022-07-11
Check has content and display proof section

T-Damer:
Check if the proof section has content and display it


src/components/ProofSection.tsx
```diff
@@ -5,6 +5,7 @@ import Section from 'components/Section'
 export default function ({
   title,
   children,
-}: ChildrenProp & { title?: ComponentChildren }) {
-  return <Section title={title}>{children}</Section>
+  hasContent = true,
```
T-Damer, line 8: `true` by default because `emailProof` section always has content by default

src/components/proofs/ProofList.tsx
```diff
@@ -21,47 +21,50 @@ export default function () {
...
             )}
-            {nothingToGenerate && (
-              <HintCard small>
-                <HintText bold center>
-                  No NFTs to proof
-                </HintText>
-              </HintCard>
-            )}
           </ProofSection>
+          {nothingToGenerate && (
```
T-Damer, line 61: It was repeatable code, so I extracted it, we can actually display `No mainnet NFTs to proof` and `No goerli NFTs to proof`, but I think we don't need this

MixailE (approved):

PR #368, opened 2022-07-12
add ERC721ProofSection

MixailE:


src/components/proofs/ERC721ProofSection.tsx
```diff
@@ -0,0 +1,74 @@
...
+  network,
+}: {
+  account: string
+  network: Network
+}) {
+  return (
+    <Suspense
+      fallback={
+        <ProofSection title={<BodyText>{network} NFTs</BodyText>}>
+          {' '}
```
T-Damer, line 62: I guess we don't need this

src/helpers/getOwnedERC721.ts
```diff
@@ -72,12 +72,14 @@ export default async function (
     const value = tokenId.toString()
 
     if (!addressToTokenIds[address]) {
-      addressToTokenIds[address] = new Set([value])
-    } else if (addressToTokenIds[address].has(value)) {
-      addressToTokenIds[address].delete(value)
-      if (!addressToTokenIds[address].size) delete addressToTokenIds[address]
+      addressToTokenIds[address] = [value]
+    } else if (addressToTokenIds[address].includes(value)) {
```
T-Damer, line 76: Maybe we can un-nest this logic?

src/icons/Erc721Badge.tsx
```diff
@@ -27,7 +28,9 @@ const pathClasses = classnames(
 )
 
 export default function () {
-  return Math.random() > 0.5 ? (
+  const random = useMemo(() => Math.random() > 0.5, [])
```
T-Damer, line 31: So we should remove it now

src/components/proofs/ERC721ProofSection.tsx
T-Damer, line 62: I think we don't need this, are we?

src/components/badges/ListTitle.tsx
```diff
@@ -4,12 +4,14 @@ import { useSnapshot } from 'valtio'
...
 import SealCredStore from 'stores/SealCredStore'
+import useContractsOwned from 'hooks/useContractsOwned'
 import useProofsAvailableToMint from 'hooks/useProofsAvailableToMint'
 
 function ListTitleSuspended() {
   const { derivativeContracts = [] } = useSnapshot(SealCredStore)
   const proofsAvailableToMint = useProofsAvailableToMint()
-  const { contractsOwned } = useSnapshot(GoerliContractsStore)
+  const contractsOwned = useContractsOwned(GoerliContractsStore)
+  // const { contractsOwned } = useSnapshot(GoerliContractsStore)
```
T-Damer, line 14: Let's remove the comment

src/icons/Erc721Badge.tsx
smyzin: Returned as it was

src/stores/ContractsStore.ts
```diff
@@ -13,9 +13,10 @@ import PersistableStore from 'stores/persistence/PersistableStore'
 import WalletStore from 'stores/WalletStore'
 import transformObjectValues from 'helpers/transformObjectValues'
 
-class ContractsStore extends PersistableStore {
+export class ContractsStore extends PersistableStore {
```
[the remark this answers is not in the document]
MixailE: I have added export for`useContractsOwned`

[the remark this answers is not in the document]
smyzin: Removed

src/helpers/getOwnedERC721.ts
smyzin: Like in the updates?

src/stores/ContractsStore.ts
```diff
@@ -16,6 +16,7 @@ import transformObjectValues from 'helpers/transformObjectValues'
 class ContractsStore extends PersistableStore {
   connectedAccounts: { [account: string]: ContractSynchronizer } = {}
   currentBlock?: number
+  addressToTokenIds?: Promise<{ [address: string]: string[] } | undefined>
```
T-Damer, line 19: Should we do both `?` and `undefined` here? Like I understand, that the value itself might be `undefined` and promise might return `undefined`, do we need both?

src/stores/ContractsStore.ts
```diff
@@ -59,10 +59,17 @@ class ContractsStore extends PersistableStore {
...
 
-    await this.connectedAccounts[WalletStore.account].getOwnedERC721(
-      this.currentBlock,
-      this.network
-    )
+    if (!this.addressToTokenIds || accountChange) {
+      this.addressToTokenIds = this.connectedAccounts[
+        WalletStore.account
+      ].syncAddressToTokenIds(this.currentBlock, this.network)
+    } else {
```
T-Damer, line 66: Let's not do `else`, let's do return in the upper `if`

src/stores/ContractsStore.ts
smyzin: I may be undefined or it may be Promise with undefined as I understood. So looks fine

src/stores/ContractsStore.ts
smyzin: Resolved

PR #375, opened 2022-07-13
Added margins for hint cards where needed

T-Damer:
- We can't use `space-y` in wrapping `<Card />` component, because it will break the `<Fade />` spacing for scrollbar

image
image


smyzin (approved):
It looks good, the main thing is to double-check all pages/components where `HintCard` is used

PR #376, opened 2022-07-13
Add FF scroll space

T-Damer:
- Used hacky way with borderRight


src/components/Scrollbar.tsx
```diff
@@ -22,17 +22,28 @@ export default function ({
   const hasWebKit = useHasWebkit()
   const { isOverflow, wrapperRef } = useIsOverflow()
 
+  const ffBorderRight =
+    isOverflow && !hasWebKit ? '1.5rem solid #0d0030' : 'none'
+  const wrapperRight = isOverflow && hasWebKit ? '0.7rem' : '0rem'
```
smyzin, line 27: I wouldn't say I like numbers and hex color instead of tailwind color here. But if there is no choice to do that, okay, then

smyzin (approved):
Looks good except one comment

PR #380, opened 2022-07-15
Fix button behavior

smyzin:
- added common disabled var
- fixed styles according to the design
- fixed button's behavior

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -35,11 +35,10 @@ import Loading from 'icons/Loading'
 import React from 'react'
 
 const commonClasses = (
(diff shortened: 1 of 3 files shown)


src/components/Button.tsx
```diff
@@ -160,27 +163,25 @@ export default function ({
...
         fullWidth,
+        available,
         center,
         type,
-        loading,
-        disabled,
         small,
       })}
-      disabled={loading || disabled}
+      disabled={available}
```
T-Damer, line 177: `!avaliable`?

smyzin: Fixed, thx

PR #383, opened 2022-07-22
remove throttling from navbar

upacyxou:
- throttled scroll removed


tuansydau:
The changes look good - I wasn't able to find a use for throttling either

MixailE (approved):

PR #384, opened 2022-07-25
add noopener and fix EnsAddress fallback

MixailE:


src/components/navbar/Wallet.tsx
```diff
@@ -73,12 +73,17 @@ export default function () {
...
-            window
-              .open(getEtherscanAddressUrl(account, Network.Goerli), '_blank')
-              ?.focus()
-          } else {
-            await WalletStore.connect(true)
+            const newWindow = window.open(
+              getEtherscanAddressUrl(account, Network.Goerli),
+              '_blank'
+            )
+            if (newWindow) {
```
T-Damer, line 80: Maybe `if (!newWindow) return`

src/components/EnsAddress.tsx
```diff
@@ -13,27 +13,33 @@ interface EnsAddressProps {
 function EnsAddressSuspended({
   address,
   truncate,
-}: EnsAddressProps & { truncate?: boolean }) {
+  truncateSize,
```
T-Damer, line 16: If you're using `truncateSize`, this means that you want to truncate, so you don't need two props, I guess

T-Damer (approved):

src/components/EnsAddress.tsx
smyzin: Agreed, otherwise looks good

MixailE: This is not true, `truncate` is needed for `ensName` and only when the screen size is large :)

smyzin (approved):
Lgtm

PR #391, opened 2022-07-27
refactor ProofStore

MixailE:


src/stores/ProofStore.ts
```diff
@@ -121,7 +121,7 @@ class ProofStore extends PersistableStore {
       const balanceSignature = await requestBalanceAttestation(
```
T-Damer, line 121: Let's remove double check of `account` above

T-Damer (approved):
cool

PR #394, opened 2022-07-27
Use <a> instead of window.open

T-Damer:
- Explain what this PR changes
- in one or more
- bullet points

### DIFF
--- a/src/components/navbar/Wallet.tsx
+++ b/src/components/navbar/Wallet.tsx
@@ -19,7 +19,6 @@ import classnames, {
   width,
 } from 'classnames/tailwind'
 import getEtherscanAddressUrl from 'helpers/getEtherscanAddressUrl'
-import openNewTabAndFocus from 'helpers/openNewTabAndFocus'
 import useBreakpoints from 'hooks/useBreakpoints'
 
 const walletContainer = classnames(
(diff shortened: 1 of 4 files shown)


src/components/navbar/Wallet.tsx
```diff
@@ -69,28 +68,33 @@ export default function () {
...
-              'No wallet connected'
-            )}
-          </AccentText>
-        </div>
-        <div className={width('w-fit')}>
-          <SealWallet connected={!!account} />
+        <div
+          className={accountLinkContainer}
+          onClick={async () => {
+            if (!account) return await WalletStore.connect(true)
```
MixailE, line 79: Can we simplify it like that `account && WalletStore.connect(true)`?

MixailE (approved):

src/components/navbar/Wallet.tsx
```diff
@@ -69,28 +68,33 @@ export default function () {
...
       )}
-      <div
-        className={accountLinkContainer}
-        onClick={async () => {
-          if (!account) return await WalletStore.connect(true)
-          openNewTabAndFocus(getEtherscanAddressUrl(account, Network.Goerli))
-        }}
+      <a
+        href={
+          account ? getEtherscanAddressUrl(account, Network.Goerli) : undefined
```
smyzin, line 73: Don't we need to place it inside `div.accountLinkContainer`?

smyzin (approved):
Lgtm

src/components/navbar/Wallet.tsx
T-Damer: Ah, yes

src/components/navbar/Wallet.tsx
T-Damer: This will break the layout

PR #395, opened 2022-07-27
Remove NULL=\u0000 from contract names

T-Damer:
- Solves layout problems on windows-chrome


smyzin (approved):
Strange but looks okay

PR #398, opened 2022-08-01
Add change network button if user uses different network

T-Damer:
- We can probably upgrade the `networkPick()` to include all the needed info to use it in the future


src/stores/WalletStore.ts
```diff
@@ -95,6 +102,47 @@ class WalletStore extends PersistableStore {
...
+        : `https://${network}.etherscan.io/`
+
+    try {
+      await provider.send('wallet_switchEthereumChain', [{ chainId }])
+      this.needNetworkChange = false
+    } catch (error) {
+      const code = serializeError(error).code
+      if (code !== 4902) return
+
+      await provider.send('wallet_addEthereumChain', [
```
T-Damer, line 130: Works if the user doesn't have our network (bnb wallet, for example)

.env.sample
```diff
@@ -1,6 +1,7 @@
 VITE_ENCRYPT_KEY=000000000000000000000000000000000000000000
 VITE_APP_NAME=Seal-Cred
-VITE_ETH_NETWORK=rinkeby
+VITE_ETH_NETWORK=goerli
+VITE_CHAIN_ID=5
```
smyzin, line 4: Can we escape setting `chain_id`?

src/stores/WalletStore.ts
```diff
@@ -95,6 +102,48 @@ class WalletStore extends PersistableStore {
...
+        {
+          chainId,
+          rpcUrls: [`https://${network}.infura.io/v3/`],
+          chainName: `${networkName} Test Network`,
+          nativeCurrency: {
+            name: currency,
+            symbol: currency,
+          },
+          blockExplorerUrls: [blockExplorerUrl],
+        },
```
smyzin, line 141: Maybe we can pass it to a separate file? What do you think?

src/stores/WalletStore.ts
```diff
@@ -95,6 +102,48 @@ class WalletStore extends PersistableStore {
...
+      network === 'mainnet'
+        ? 'https://etherscan.io/'
+        : `https://${network}.etherscan.io/`
+
+    try {
+      await provider.send('wallet_switchEthereumChain', [{ chainId }])
+      this.needNetworkChange = false
+    } catch (error) {
+      const code = serializeError(error).code
+      if (code !== 4902) return
```
smyzin, line 129: Don't we need to set `needNetworkChange` to `true` here?

smyzin (approved):
Left several comments, otherwise looks good

.env.sample
T-Damer: Oh, forgot to remove it

src/stores/WalletStore.ts
T-Damer: it's already set in `#L110`

PR #399, opened 2022-08-01
Very cool animation for lists done with two lines on code

T-Damer:


src/components/proofs/ContractListContainer.tsx
```diff
@@ -8,5 +10,10 @@ const container = classnames(
 )
 
 export default function ({ children }: ChildrenProp) {
-  return <div className={container}>{children}</div>
+  const listRef = useAutoAnimate({})
+  return (
+    <div className={container} ref={listRef as Ref<HTMLDivElement>}>
```
smyzin, line 15: Can't we do `const listRef = useAutoAnimate<HTMLDivElement>({})`?

src/components/badges/BadgeSection.tsx
```diff
@@ -30,10 +31,11 @@ export default function ({
   onMinted?: () => void
 }) {
   if (minted.length === 0 && proofs.length === 0) return null
+  const listRef = useAutoAnimate({})
```
smyzin, line 34: Also, is it possible to set it as `useAutoAnimate<HTMLDivElement>`?

T-Damer: It doesnt eat typings

src/components/proofs/ContractListContainer.tsx
T-Damer: No, we can't, it doesn't accept type notations

smyzin (approved):
Lgtm

src/components/badges/BadgeSection.tsx
```diff
@@ -30,10 +31,11 @@ export default function ({
   onMinted?: () => void
 }) {
   if (minted.length === 0 && proofs.length === 0) return null
```
MixailE, line 33: This return will not lead to errors?

T-Damer: Fixed it

PR #400, opened 2022-08-01
replaced user signatures r2 and s2 with 64-bit random numbers

tuansydau:
## What 
* Replaced `r2` and `s2` in ERC721Proof.ts with random numbers as opposed to user signatures.

### DIFF
--- a/src/helpers/ERC721Proof.ts
+++ b/src/helpers/ERC721Proof.ts
@@ -106,10 +106,11 @@ export default class ERC721Proof
   async generateInputs(
     ownershipSignature: Signature,
     balanceSignature: BalanceSignature,
-    nullifierSignature: string,
     eddsaPublicKey: PublicKey
   ) {
-    const { r: r2, s: s2 } = utils.splitSignature(nullifierSignature)
+    const randomNumbersArray = crypto.getRandomValues(new BigUint64Array(2))
+    const r2 = randomNumbersArray[0].toString()
+    const s2 = randomNumbersArray[1].toString()
     const addressInputs = await this.inputsForSignature(
       eddsaPublicKey,
       ownershipSignature,
@@ -131,13 +132,11 @@ export default class ERC721Proof
   async build(
     ownershipSignature: Signature,
     balanceSignature: BalanceSignature,
-    nullifierSignature: string,
     eddsaPublicKey: PublicKey
   ) {
     const inputs = await this.generateInputs(
       ownershipSignature,
       balanceSignature,
-      nullifierSignature,
       eddsaPublicKey
     )
     this.result = await snarkjs.groth16.fullProve(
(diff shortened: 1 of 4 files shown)


src/helpers/ERC721Proof.ts
```diff
@@ -106,10 +106,11 @@ export default class ERC721Proof
   async generateInputs(
     ownershipSignature: Signature,
     balanceSignature: BalanceSignature,
-    nullifierSignature: string,
     eddsaPublicKey: PublicKey
   ) {
-    const { r: r2, s: s2 } = utils.splitSignature(nullifierSignature)
+    const randomNumbersArray = crypto.getRandomValues(new BigUint64Array(2))
```
MixailE, line 111: How about doing it like this? ```const [r2, s2] = crypto.getRandomValues(new BigUint64Array(2))```

src/helpers/EmailProof.ts
```diff
@@ -60,15 +60,16 @@ export default class EmailProof extends BaseProof implements EmailProofSchema {
...
-    nullifierSignature: string
+    eddsaPublicKey: PublicKey
   ) {
     const maxDomainLength = 90
     const messageUInt8 = this.padZeroesOnRightUint8(
       utils.toUtf8Bytes(domain),
       maxDomainLength
     )
-    const { r: r2, s: s2 } = utils.splitSignature(nullifierSignature)
+    const randomNumbersArray = crypto.getRandomValues(new BigUint64Array(2))
```
MixailE, line 70: How about doing it like this? ```const [r2, s2] = crypto.getRandomValues(new BigUint64Array(2))```

MixailE: I also noticed that we have a file that isn't used anywhere `unpackECDSASignature.ts`

src/helpers/ERC721Proof.ts
tuansydau: This would be better, but returns r2 and s2 in the form "{xxxx}n" since it's a BigInt array - I'd have to cast to a string after wards to get rid of the n anyways, which would look equally as messy 

Ie, it'd look like something pretty similar to what we have now, unless there's a way to cast the BigInt values to strings in-place
```
var [r2, s2] = crypto.getRandomValues(new BigUint64Array(2))
var r2 = r2.toString()
var s2 = s2.toString()
```

src/helpers/EmailProof.ts
tuansydau: > How about doing it like this? `const [r2, s2] = crypto.getRandomValues(new BigUint64Array(2))`

Same issue as above, though I wish that this worked since it's a really clean solution.

tuansydau: > I also noticed that we have a file that isn't used anywhere `unpackECDSASignature.ts`

Deleted the file, I didn't find any references in `seal-cred-frontend`. Thanks

tuansydau: Nvm it works

src/helpers/ERC721Proof.ts
tuansydau: The nullifier takes bigInt, disregard what I said before. Thanks!

backmeupplz (approved):

PR #403, opened 2022-08-02
Add a back arrow during the token step of the email flow

upacyxou:
- back button during token step added

### DIFF
--- a/src/components/proofs/EmailProof.tsx
+++ b/src/components/proofs/EmailProof.tsx
@@ -68,7 +68,7 @@ const proofLineContainer = classnames(
 
 const emailTitleLeft = classnames(
   display('flex'),
-  space('space-x-2'),
+  space('space-x-3'),
   alignItems('items-center')
 )
 
@@ -96,10 +96,6 @@ export default function () {
     setGenerationStarted(false)
   }
 
-  function onGenerationStarted() {
-    setGenerationStarted(true)
-  }
-
   function jumpToToken() {
     setError(undefined)
     setDomain(emailDomain)
@@ -121,9 +117,7 @@ export default function () {
                   setDomain('')
                 }}
               >
-                <div>
-                  <SimpleArrow />
-                </div>
+                <SimpleArrow />
               </Button>
             )}
             <ProofText>Work email</ProofText>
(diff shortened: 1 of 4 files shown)


src/components/proofs/EmailProof.tsx
```diff
@@ -83,13 +85,19 @@ export default function () {
...
   const { emailDomain } = useSnapshot(EmailDomainStore)
   const [containerRef] = useAutoAnimate<HTMLDivElement>()
 
   function onCreate() {
     setOpen(false)
     setDomain('')
+    setGenerationStarted(false)
+  }
+
+  function onGenerationStarted() {
```
T-Damer, line 99: We use it only once, I think we shouldn't extract it into separate function

src/components/proofs/EmailProofForm.tsx
```diff
@@ -53,6 +55,10 @@ export default function ({
     if (!checkDomainToken(secret))
       return onError('This is an invalid token. Please try again.')
 
+    if (onGenerationStarted) {
```
T-Damer, line 58: Let's remove `{}`, do:
```TS
if (onGenerationStarted) onGenerationStarted()
```
Visually should look a little better :)

src/icons/SimpleArrow.tsx
```diff
@@ -0,0 +1,19 @@
+export default function () {
```
T-Damer, line 1: Let's use `Arrow.tsx` with prop `plain` or `white`, simply change the colors inside the `svg`

src/components/proofs/EmailProof.tsx
```diff
@@ -105,6 +113,19 @@ export default function () {
...
           <div className={emailTitleLeft}>
+            {domain && (
+              <Button
+                disabled={generationStarted}
+                type={'tertiary'}
+                onClick={() => {
+                  setDomain('')
+                }}
+              >
+                <div>
```
T-Damer, line 124: I think we don't need this wrapper

T-Damer: Only if you need additional `margin-right` here maybe

T-Damer (approved):

PR #406, opened 2022-08-03
Update ledgers

smyzin:
- updated packages
- updated ledgers


src/helpers/contracts/getLedgerRecord.ts
```diff
@@ -0,0 +1,6 @@
+export default function (original: string, derivative: string) {
```
upacyxou, line 1: this function and `getEmailLedgerRecord.ts` from above looks very similar, do we actually need two different functions for this?

smyzin: Updated. Thank you!

src/helpers/contracts/getEmailLedger.ts
```diff
@@ -1,18 +1,15 @@
 import { SCEmailLedger } from '@big-whale-labs/seal-cred-ledger-contract'
 import EmailLedger from 'models/EmailLedger'
-import getEmailLedgerRecord from 'helpers/contracts/getEmailLedgerRecord'
+import getLedgerRecord from 'helpers/contracts/getLedgerRecord'
 
 export default async function (ledger: SCEmailLedger): Promise<EmailLedger> {
-  const eventsFilter = ledger.filters.CreateDerivativeContract()
+  const eventsFilter = ledger.filters.CreateDerivative()
```
T-Damer, line 6: Seems like a repetitive code, I'll extract this into function

src/models/EmailLedger.ts
```diff
@@ -1,6 +1,6 @@
 export default interface SCEmailLedger {
-  [domain: string]: {
+  [originalContract: string]: {
+    originalContract: string
```
T-Damer, line 3: Why do we need to map `originalContract` to `originalContract` if we can use the `key`?

T-Damer (approved):
lgtm

PR #409, opened 2022-08-06
add calculation for maxFeePerGas and minMaxPriorityFeePerGas

MixailE:


src/helpers/createEmailBadge.ts
```diff
@@ -11,9 +11,14 @@ function createContract(provider: Web3Provider) {
   )
 }
 
-export default async function (provider: Web3Provider, proof: EmailProof) {
+export default async function (
+  provider: Web3Provider,
+  proof: EmailProof,
+  maxFeePerGas?: string
+) {
+  if (!proof.result) throw new Error('Invalid proof')
```
T-Damer, line 19: Looks like this is the same, as in `createExternalERC721Badge.ts`, maybe we should refactor it?

backmeupplz (approved):

PR #410, opened 2022-08-08
Fix footer link and make them scroll to top

smyzin:
- fixed footer links
- created Scroll component


src/components/Text.tsx
```diff
@@ -266,24 +276,33 @@ export function LinkText({
...
-  fontWeight('font-semibold'),
-  textDecoration('no-underline', 'hover:underline'),
-  textColor('text-formal-accent', 'hover:text-accent')
-)
-export function FooterlLink({
+const footerLink = (active?: boolean) =>
+  classnames(
+    fontSize('text-sm'),
+    fontWeight('font-semibold'),
+    textDecoration(active ? 'underline' : 'no-underline', 'hover:underline'),
```
T-Damer, line 283: Let's do object here

src/components/Text.tsx
```diff
@@ -266,24 +276,33 @@ export function LinkText({
...
+    textColor(
+      active ? 'text-accent' : 'text-formal-accent',
```
T-Damer, line 285: Let's do object here

T-Damer (approved):

src/components/Text.tsx
smyzin: Fixed

src/components/Text.tsx
smyzin: Fixed

PR #411, opened 2022-08-08
Current minting will display loading, while others will be blocked

T-Damer:
- Added `mintLoading` to `WalletStore` to share state across button
- Added `[localLoading]` to mint button to display loading state only for current minting proccess

### DIFF
--- a/src/components/badges/BadgeBlock.tsx
+++ b/src/components/badges/BadgeBlock.tsx
@@ -24,13 +24,13 @@ function Badge({
   onMintFailed?: (minted?: MintedToken[]) => void
 }) {
   const { account, mintLoading } = useSnapshot(WalletStore)
-  const [localLoading, setLocalLoading] = useState(false)
+  const [loading, setLoading] = useState(false)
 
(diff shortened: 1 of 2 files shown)


src/components/badges/BadgeBlock.tsx
```diff
@@ -23,14 +23,14 @@ function Badge({
...
+  const { account, mintLoading } = useSnapshot(WalletStore)
+  const [localLoading, setLocalLoading] = useState(false)
 
   const isEmailProof = proof instanceof EmailProof
 
   const checkProofAndMint = async () => {
-    setLoading(true)
-
+    WalletStore.mintLoading = true
+    setLocalLoading(true)
```
smyzin, line 33: `Local Loading` looks weird a little. Maybe it's better to call it `isBadgeMinting` or something like that?

smyzin (approved):
Looks good to me

src/components/badges/BadgeBlock.tsx
T-Damer: Maybe let's leave `loading`, like it was

PR #414, opened 2022-08-09
Relay provider back to scene

T-Damer:


src/helpers/providers/relayProvider.ts
```diff
@@ -1,27 +1,9 @@
 import { Eip1193Bridge } from '@ethersproject/experimental'
-import { PingFilter } from '@opengsn/common'
 import { RelayProvider } from '@opengsn/provider'
 import { Web3Provider } from '@ethersproject/providers'
 import { WrapBridge } from '@opengsn/provider/dist/WrapContract'
 import env from 'helpers/env'
 
-const GasPricePingFilter: PingFilter = (
```
upacyxou, line 8: I don't think that we need to remove this, all of my local test was successful with this filter

upacyxou:
otherwise lgtm

upacyxou (changes requested):

upacyxou (approved):

vite.config.ts
```diff
@@ -10,6 +10,7 @@ import { NodeModulesPolyfillPlugin } from '@esbuild-plugins/node-modules-polyfil
 
 export default defineConfig({
   plugins: [preact(), tsconfigPaths()],
+  resolve: { alias: { assert: 'assert-browserify' } },
```
backmeupplz, line 13: We should check if removing `assert` from `inject` below breaks proof generation or minting

backmeupplz (changes requested):

backmeupplz (approved):

PR #419, opened 2022-08-09
fixed the footer position on all pages

tuansydau:
## What
* Fixes the footer to be at the bottom of the page regardless of screen size
* Passes `yarn preview`
## Demo
[Footer on a large screen](https://user-images.githubusercontent.com/26176104/183980611-c30d3d60-ab80-4584-ae95-20e0ed269bbd.png)
[Footer on a 1080p screen](https://user-images.githubusercontent.com/26176104/183980657-8f2dc6ac-ba9f-411d-a7cd-36765fd31ce7.png)


tuansydau:
Updated this PR to use flexboxes instead of the cursed `index.css` magic. New demo screenshots available in description

src/App.tsx
```diff
@@ -9,58 +9,68 @@ import Privacy from 'pages/Privacy'
...
-                  element={<LazyComponent lazyImported={<Privacy />} />}
-                />
-                <Route
-                  path="*"
-                  element={<LazyComponent lazyImported={<NotFound />} />}
-                />
-              </Routes>
-              <Footer />
-            </div>
+            <Routes>
```
T-Damer, line 43: Are you sure that removing `space-y-4` doesn't break anything?

src/components/Footer.tsx
```diff
@@ -38,43 +38,45 @@ const footerLogo = classnames(
...
+            <FooterLink internal url="/terms">
+              Terms of service
+            </FooterLink>
+            <Delimiter />
+            <FooterLink internal url="/privacy">
+              Privacy policy
+            </FooterLink>
+          </div>
+          {!md && (
+            <>
```
T-Damer, line 67: Hm, probably we can remove this `<></>`, because only one `<div>...</div>` is returned here

src/components/Footer.tsx
```diff
@@ -38,43 +38,45 @@ const footerLogo = classnames(
...
-          </FooterLink>
-          <Delimiter />
-          <FooterLink internal url="/terms">
-            Terms of service
-          </FooterLink>
-          <Delimiter />
-          <FooterLink internal url="/privacy">
-            Privacy policy
-          </FooterLink>
+      <div className={footerContainer}>
```
T-Damer, line 48: You wrapped all footer as I can see, you can remove `<></>` now

src/components/Footer.tsx
```diff
@@ -38,43 +38,45 @@ const footerLogo = classnames(
   flexDirection('flex-row'),
   space('space-x-4')
 )
+const footerContainer = classnames(inset('bottom-0', 'left-0', 'right-0'))
```
T-Damer, line 41: Have you tried doing it without `right-0` and `left-0`? Maybe you can use only `right-0`/'left-0` without each other?

T-Damer: As I know you can't position [`position: static`](https://developer.mozilla.org/en-US/docs/Web/CSS/position) (which is default for all DOM elements) with `inset, top, bottom etc...`, how does this work?

src/App.tsx
tuansydau: Oh I forgot to revive that change, thanks for the heads up

src/components/Footer.tsx
tuansydau: This was an artifact, got rid of it completely thanks for spotting it

T-Damer (approved):

PR #422, opened 2022-08-10
make animation of closing announcement smooth

upacyxou:
- smooth animation adde

Now transition goes on `height`, but not on `translate-y`, so the announcement is moving with DOM and it looks way smoother

https://user-images.githubusercontent.com/57546893/183931961-4fe7223e-4def-497c-b202-7e491aca5d4c.mp4


src/components/Announcement.tsx
```diff
@@ -12,27 +12,36 @@ import classnames, {
...
+    padding(animate ? 'px-0' : 'px-6', animate ? 'py-0' : 'py-4'),
     display('flex'),
     alignItems('items-center'),
     justifyContent('justify-center'),
     fontWeight('font-bold'),
     fontSize('text-sm'),
     lineHeight('leading-5'),
-    transitionProperty('transition-transform'),
-    translate(animate ? '-translate-y-full' : undefined)
+    height(animate ? 'h-0' : undefined),
```
T-Damer, line 42: Same here `{'h-0': animate}`

src/components/Announcement.tsx
```diff
@@ -12,27 +12,36 @@ import classnames, {
...
+    transitionProperty('transition-height', 'transition-padding'),
```
T-Damer, line 43: Better do `transition-all`, take a look here: https://developer.mozilla.org/en-US/docs/Web/CSS/transition-property

T-Damer: When you specify multiple transitions from tailwind, you specify many things like timing-function, duration: https://tailwindcss.com/docs/transition-property, they will just exist as dead code in CSS

tailwind.config.js
```diff
@@ -139,11 +139,15 @@ module.exports = {
...
         '-5.5': '-0.35rem',
         '-7': '-0.45rem',
-        '-full': '-100%',
       },
       backgroundImage: {
         noise: 'url("/img/noise50.png")',
       },
+      transitionProperty: {
+        height: 'height',
+        'line-height': 'line-height',
```
smyzin, line 148: Do we need it?

src/components/Announcement.tsx
```diff
@@ -51,18 +60,26 @@ export default function ({ redirectTo }: { redirectTo: string }) {
   const onPage = location.pathname === redirectTo
 
   return (
-    <div id="bottom-bar" className={announceWrapper(animate)}>
+    <div className={announceWrapper(animate)}>
```
T-Damer, line 63: Why we needed an `id` here?

src/components/Announcement.tsx
```diff
@@ -51,18 +60,26 @@ export default function ({ redirectTo }: { redirectTo: string }) {
...
-          {!onPage
-            ? 'Get started'
-            : account
-            ? 'Get started below.'
-            : 'Connect wallet to get started.'}
-        </AccentText>
-      </LinkText>
+      <div
+        className={classnames(
+          opacity(animate ? 'opacity-0' : undefined),
```
T-Damer, line 67: opacity-100 is default, let's do `{'opacity-0': animate}`

src/components/Announcement.tsx
upacyxou: Don't have any idea, it's not used anywhere in the code

src/components/Announcement.tsx
T-Damer: And I think we need to extract those classnames into separate `const`

src/components/Announcement.tsx
```diff
@@ -51,18 +60,26 @@ export default function ({ redirectTo }: { redirectTo: string }) {
   const onPage = location.pathname === redirectTo
 
   return (
-    <div id="bottom-bar" className={announceWrapper(animate)}>
```
smyzin, line 54: We don't need it anymore?

src/components/Announcement.tsx
```diff
@@ -51,18 +60,26 @@ export default function ({ redirectTo }: { redirectTo: string }) {
...
+        className={classnames(
+          opacity(animate ? 'opacity-0' : undefined),
+          transitionProperty('transition-opacity'),
+          commonTransition
+        )}
+      >
+        <LinkText url={redirectTo}>
+          <AccentText small bold color="text-formal-accent">
+            Now introducing zk proofs for your work email!{' '}
+            {!onPage
```
T-Damer, line 75: Let's correct the logic here:
```TS
onPage ? account ? 'Get started below.' : 'Connect wallet to get started' : 'Get started'
```
I think it's easier to read, you can extract central block like:
```TS
const hasAccount = account ? account ? 'Get started below.' : 'Connect wallet to get started'
onPage ? hasAccount : 'Get started'
```

T-Damer (commented):
let's clean up a little :)

src/components/Announcement.tsx
upacyxou: It's not used in the code at all

src/components/Announcement.tsx
upacyxou: Got it, thank you

src/components/Announcement.tsx
upacyxou: Thank you! Fixed

tailwind.config.js
upacyxou: Not anymore, fixed, thank you

backmeupplz (approved):

PR #431, opened 2022-08-11
Fix scrollbar

smyzin:
- added `margin-y-auto` to the main content
- changed the position of blocks

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -12,7 +12,7 @@ import Terms from 'pages/Terms'
 import classnames, {
   display,
   flexDirection,
-  justifyContent,
+  margin,
   minHeight,
   space,
 } from 'classnames/tailwind'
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -27,22 +27,20 @@ const pageContainer = classnames(
   space('space-y-6', 'sm:space-y-0'),
   display('flex'),
   flexDirection('flex-col'),
-  minHeight('min-h-screen'),
-  justifyContent('justify-between')
+  minHeight('min-h-screen')
 )
+const mainContent = classnames(space('space-y-4'), margin('sm:!my-auto'))
```
smyzin, line 32: I know that using `important` is wrong, but it's the only way how to overwrite `sm:space-y-0` properties and make the space between header/main content/footer responsive

PR #433, opened 2022-08-12
add transformRelayErrorMessage

MixailE:


src/helpers/providers/relayProvider.ts
```diff
@@ -4,6 +4,18 @@ import { Web3Provider } from '@ethersproject/providers'
 import { WrapBridge } from '@opengsn/provider/dist/WrapContract'
 import env from 'helpers/env'
 
+export function transformRelayErrorMessage(message: string) {
```
T-Damer, line 7: I think we should put it into `handleError`, because basically it's parser

src/helpers/handleError.ts
```diff
@@ -29,5 +30,5 @@ export default function (error: unknown) {
   }
   if (!displayedError) displayedError = ErrorList.unknown
 
-  toast.error(displayedError)
+  toast.error(transformRelayErrorMessage(displayedError))
```
T-Damer, line 33: Can we parse error before toasting and reassign `displayedError`? If we'll need another parser to connect, we'll need to wrap this message again, that's not good

T-Damer (approved):

PR #434, opened 2022-08-12
Make gradient buttons

smyzin:
- made new component `GradientButton`
- fixed existing component `Button`

Снимок экрана 2022-08-12 в 15 16 39

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -39,7 +39,8 @@ const commonClasses = (
   fullWidth?: boolean,
   center?: boolean,
   available?: boolean,
-  small?: boolean
+  small?: boolean,
+  gradientFont?: boolean
 ) =>
   classnames(
     display('flex'),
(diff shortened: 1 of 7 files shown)


src/components/Button.tsx
```diff
@@ -107,14 +113,17 @@ const colorClasses = (type: ButtonType, available?: boolean) =>
...
           borderWidth('border'),
           borderRadius('rounded-full'),
-          borderColor(
-            'border-secondary',
-            'hover:border-secondary',
-            'active:border-secondary'
-          ),
+          borderColor({
+            'border-transparent': gradientFont,
+            'border-secondary': !gradientFont,
```
T-Damer, line 118: I just think, that code goes through here and does conversion each time you do `!something`, maybe we should convert it in function body and pass as variable, wdyt?

T-Damer (approved):

T-Damer (approved):

PR #435, opened 2022-08-12
Fix navbar with scroll

smyzin:
- fixed navbar (now it's scrolling with the page)
- updated page layout

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -31,7 +31,7 @@ const pageContainer = classnames(
 const bodyContainer = classnames(
   display('flex'),
   flexDirection('flex-col'),
-  margin('my-6', 'md:mt-0', 'md:mb-auto', 'mx-auto')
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -24,25 +23,26 @@ const Main = lazy(() => import('pages/Main'))
...
   flexDirection('flex-col'),
-  minHeight('min-h-screen'),
-  justifyContent('justify-between')
+  minHeight('min-h-screen')
+)
+
+const bodyContainer = classnames(
+  display('flex'),
+  flexDirection('flex-col'),
+  margin('my-6', 'md:mt-0', 'md:mb-auto', 'mx-auto')
```
T-Damer, line 34: Let's move `mx-auto` to the left part of this, because you have left part for `mobiles` and right part for `desktops`

T-Damer (approved):

src/App.tsx
smyzin: Did it

PR #436, opened 2022-08-12
Implement logo instead of generating it in canvas

smyzin:
- removed image from options
- added our logo as image and styled it
- rechecked it in MetaMask

image

### DIFF
--- a/src/components/QRCode.tsx
+++ b/src/components/QRCode.tsx
@@ -2,9 +2,16 @@ import { useLayoutEffect, useRef } from 'react'
 import QRCodeStyling, { Options } from 'qr-code-styling'
 import classNamesToString from 'helpers/classNamesToString'
 import classnames, {
+  alignItems,
+  backgroundColor,
   borderRadius,
+  display,
+  inset,
+  justifyContent,
   minWidth,
   overflow,
+  position,
+  width,
 } from 'classnames/tailwind'
 
 interface QRCodeProps {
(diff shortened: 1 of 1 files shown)


src/components/QRCode.tsx
```diff
@@ -13,15 +20,26 @@ interface QRCodeProps {
...
+)
+const qrCodeLogo = classnames(width('w-8'), backgroundColor('bg-primary-dark'))
 
 const QRCodeOptions: Options = {
   type: 'canvas',
   margin: 10,
-  image: '/img/logo.svg',
+  width: 200,
+  height: 200,
+  image: '',
```
T-Damer, line 42: What if we don't specify `image` at all?

src/components/QRCode.tsx
```diff
@@ -54,10 +72,14 @@ export default function ({ derivativeAddress, tokenId }: QRCodeProps) {
   return (
     <a
       href={url}
-      ref={ref}
       className={classNamesToString(qrCodeContainer, 'custom-qr-code')}
       target="_blank"
       rel="noopener noreferrer"
-    />
+    >
+      <div ref={ref} />
```
T-Damer, line 79: What does this `div` do?

src/components/QRCode.tsx
smyzin: Nothing, if we didn't specify an image, it won't generate logo in the center of the QR

src/components/QRCode.tsx
smyzin: `QRCodeStyling` generates canvas with QR into that div block. It is needed because generating QR will overwrite all children in the element. So not to lose the `logo`, we need to put it near it in a separate block

T-Damer (approved):

PR #438, opened 2022-08-13
changed the width of the main cards

tuansydau:
## What 
* Changes the main app cards to have the new widths and heights from the new SealCred scalability design
* Main app cards are now responsive too 
## Demo
[Screencast from 2022-08-12 08:48:01 PM.webm](https://user-images.githubusercontent.com/26176104/184461128-edbb3ad6-fff9-4691-8a82-582975b71346.webm)


src/components/Card.tsx
```diff
@@ -74,7 +76,8 @@ const cardContainer = (
   nospace?: boolean,
   useAppStyles?: boolean,
   paddingType?: 'small' | 'normal',
-  spinner?: string
+  spinner?: string,
+  bigCard?: boolean
```
T-Damer, line 80: We have `useAppStyles`, can we use them for this `only-app` cards? We should modify that prop

src/pages/Landing.tsx
```diff
@@ -31,7 +31,8 @@ const pageBox = classnames(
   display('flex'),
   flexDirection('flex-col'),
   alignItems('items-center'),
-  margin('mb-10')
+  margin('mb-10'),
+  width('w-full')
```
T-Damer, line 35: Why do we need to change something on landing? This might affect small devices and some breakpoints/animations/etc

src/pages/Main.tsx
```diff
@@ -2,43 +2,62 @@ import Badges from 'components/badges/Badges'
...
 import useBreakpoints from 'hooks/useBreakpoints'
 
-const mainBlock = classnames(
+const cardsContainer = classnames(
   display('flex'),
   flexDirection('flex-col', 'lg:flex-row'),
   alignItems('items-center', 'lg:items-stretch'),
-  justifyContent('lg:justify-center')
+  justifyContent('lg:justify-center'),
+  margin('ml-10', 'mr-10')
```
T-Damer, line 24: `mx-10`

src/pages/Main.tsx
```diff
@@ -2,43 +2,62 @@ import Badges from 'components/badges/Badges'
...
+        <ProofsCard />
+        <CardSeparator
+          numberOfLines={3}
+          gradient="accent-to-secondary"
+          vertical={!lg}
+        />
+        <Badges />
+      </div>
+      <div className={proofContainer}>
+        {!lg && (
```
T-Damer, line 49: Let's remove this breakpoint and use tailwind `display(lg:hidden)`

src/pages/Main.tsx
```diff
@@ -2,43 +2,62 @@ import Badges from 'components/badges/Badges'
...
-        <>
-          <CardSeparator
-            numberOfLines={1}
-            gradient="secondary-to-transparent"
-            vertical
-          />
-          <ZkProofHint />
-        </>
-      )}
+    <div className={mainContainer}>
```
T-Damer, line 38: Do we really need to nest two `div`s into each other like this? Usually you can avoid doing it

src/components/Card.tsx
```diff
@@ -88,13 +91,24 @@ const cardContainer = (
       'py-5': paddingType === 'normal',
       'py-8': typeof paddingType === 'undefined',
     }),
-    width(
-      thin ? 'sm:!w-thin-card' : 'sm:w-card',
-      thin ? 'tiny:w-thin-mobile' : 'w-mobile-card',
-      { 'w-32': thin }
-    ),
+
+    bigCard
```
T-Damer, line 95: We need to simplify this logic, I think there should be a way

src/components/Card.tsx
```diff
@@ -88,13 +91,24 @@ const cardContainer = (
...
+      ? width('w-full')
+      : width(
+          thin ? 'sm:!w-thin-card' : 'sm:w-card',
+          thin ? 'tiny:w-thin-mobile' : 'w-mobile-card',
+          { 'w-32': thin }
+        ),
+    height(bigCard ? 'h-big-card' : undefined),
```
T-Damer, line 102: `height({'h-big-card': bigCard})` is preferrable

src/components/Card.tsx
```diff
@@ -88,13 +91,24 @@ const cardContainer = (
...
     margin({ 'mx-auto': !thin }, 'lg:mx-0'),
-    maxHeight({ 'sm:max-h-card': !onlyWrap, 'max-h-mobile-card': !onlyWrap }),
+    maxHeight(
```
T-Damer, line 104:
```TS
maxHeight({'max-h-big-card': bigCard, 'sm:max-h-card': !onlyWrap && !bigCard, 'max-h-mobile-card': !onlyWrap && !bigCard})
```
Please don't mix `{}` and `?:` things

src/components/Card.tsx
```diff
@@ -88,13 +91,24 @@ const cardContainer = (
...
+      bigCard
+        ? 'max-h-big-card'
+        : {
+            'sm:max-h-card': !onlyWrap,
+            'max-h-mobile-card': !onlyWrap,
+          }
+    ),
     space({ 'space-y-6': !nospace }),
     margin(spinner ? 'mt-4' : undefined),
```
T-Damer, line 113: Let's do `margin({'mt-4': spinner})`

tuansydau:
Thanks for the comments - I'll work on adding these comments in soon

src/pages/Landing.tsx
tuansydau: This fixes an error with the SuperHr component, but I'll remove this, log this in trello, and do it in a separate PR instead actually so that we can test it thoroughly

The error is that without the div taking up the whole page width, SuperHr doesn't actually span the whole screen. Also it should totally have been width('w-screen') instead

src/pages/Main.tsx
tuansydau: true

src/pages/Main.tsx
tuansydau: This mainBlock div holds two separate divs, the cardsContainer for positioning the cards, and then the proofHintContainer for positioning the ZKProofHint

src/components/Card.tsx
tuansydau: I think I've simplified it a bit, wdyt @T-Damer ?

src/components/Card.tsx
tuansydau: I was able to get rid of the bigCard property entirely

src/components/Card.tsx
```diff
@@ -88,15 +92,18 @@ const cardContainer = (
...
+      : undefined,
     margin({ 'mx-auto': !thin }, 'lg:mx-0'),
-    maxHeight({ 'sm:max-h-card': !onlyWrap, 'max-h-mobile-card': !onlyWrap }),
+    maxHeight({
+      'sm:max-h-card': !onlyWrap && !useAppStyles,
+      'max-h-mobile-card': !onlyWrap && !useAppStyles,
+    }),
     space({ 'space-y-6': !nospace }),
-    margin(spinner ? 'mt-4' : undefined),
+    margin({ 'mt-4': spinner != null }),
```
upacyxou, line 106: always use strict equal

tailwind.config.js
```diff
@@ -17,9 +17,11 @@ module.exports = {
         card: '31.75rem',
         'mobile-card': '41rem',
         'badges-list': '21.25rem',
+        'app-card': '77.75rem',
       },
       maxWidth: {
         100: '6.25rem',
+        1244: '77.75rem',
```
upacyxou, line 24: what is 1244?

src/components/Card.tsx
```diff
@@ -88,15 +92,18 @@ const cardContainer = (
       'py-5': paddingType === 'normal',
       'py-8': typeof paddingType === 'undefined',
     }),
-    width(
-      thin ? 'sm:!w-thin-card' : 'sm:w-card',
-      thin ? 'tiny:w-thin-mobile' : 'w-mobile-card',
-      { 'w-32': thin }
-    ),
+    !useAppStyles
```
upacyxou, line 95: I think that it's hard to read maybe you could split this into constsants?

src/components/proofs/Card.tsx
```diff
@@ -1,27 +1,29 @@
...
-      </Card>
-      <ZkProofHintWhenLg />
-    </CardContainer>
+    <Card
+      paddingType="normal"
+      color="accent"
+      shadow
+      useAppStyles
+      nospace
+      onlyWrap
```
smyzin, line 18: These two may lead to weird behavior. Did you test and checked everything?

tailwind.config.js
tuansydau: 1244 is the max-size for both of the appScreen cards outlined here - https://www.figma.com/file/mX1w2b1WxPn7QJ851bcy59/SealCred-May-2022?node-id=2066%3A46357

I originally had this maxWidth named as `big-cards` but Nikita said that it was a confusing name, maybe I should name it to something like `app-content`?

src/components/Card.tsx
tuansydau: Tyty

src/components/Card.tsx
tuansydau: good idea, tyty

src/components/proofs/Card.tsx
tuansydau: I've tested moderately and I haven't had any issues with any element positioning, but i'll test a bit more just to make sure

tailwind.config.js
tuansydau: Changed it to `app-content`

ohpaul (approved):

src/components/Card.tsx
```diff
@@ -88,15 +96,14 @@ const cardContainer = (
       'py-5': paddingType === 'normal',
       'py-8': typeof paddingType === 'undefined',
     }),
-    width(
-      thin ? 'sm:!w-thin-card' : 'sm:w-card',
-      thin ? 'tiny:w-thin-mobile' : 'w-mobile-card',
-      { 'w-32': thin }
-    ),
+    !useAppStyles ? (thin ? thinWidthStyles : defaultWidthStyles) : undefined,
```
T-Damer, line 99: Don't convert `!useAppStyels`, do:
```TS
useAppStyles ? undefined : (thin ? ... : ... )
```

src/components/Card.tsx
```diff
@@ -88,15 +96,14 @@ const cardContainer = (
...
     margin({ 'mx-auto': !thin }, 'lg:mx-0'),
-    maxHeight({ 'sm:max-h-card': !onlyWrap, 'max-h-mobile-card': !onlyWrap }),
+    maxHeight({
+      'sm:max-h-card': !onlyWrap && !useAppStyles,
```
T-Damer, line 102: Seems like you do `!onlyWrap && !useAppStyles` twice, lets can extract this into `const`

src/components/Card.tsx
```diff
@@ -88,15 +96,14 @@ const cardContainer = (
...
+      'max-h-mobile-card': !onlyWrap && !useAppStyles,
+    }),
     space({ 'space-y-6': !nospace }),
-    margin(spinner ? 'mt-4' : undefined),
+    margin({ 'mt-4': spinner !== null }),
```
T-Damer, line 106:
```TS
margin({'mt-4': spinner})
OR
margin({'mt-4': !!spinner})
```
Should work too

src/components/Card.tsx
T-Damer: Fewer conversions = easier reading + faster code

src/pages/Main.tsx
T-Damer: Yeah, I can see it now, thanks

src/components/Card.tsx
tuansydau: First one gets rejected by `yarn build`, but second one works :) Thanks

src/components/Card.tsx
tuansydau: Oh I like this change, thanks

src/components/Card.tsx
tuansydau: Fixed

tuansydau:
Addressed your changes @T-Damer

src/components/Card.tsx
```diff
@@ -64,9 +65,16 @@ const appStyles = classnames(
   display('flex'),
   flexDirection('flex-col'),
   alignItems('items-stretch'),
-  minHeight('sm:min-h-card', 'min-h-fit')
+  minHeight('sm:min-h-card', 'min-h-fit'),
+  maxHeight('max-h-app-card'),
```
T-Damer, line 69: Not sure if we need all 3 `height, minHeight, maxHeight`, I think we can get rid of `height`

T-Damer: Seems like mobile card is broken 
image

src/components/Card.tsx
```diff
@@ -64,9 +65,16 @@ const appStyles = classnames(
...
+  height('h-app-card'),
+  width('w-full')
```
T-Damer, line 71: As I can see, there's not that much `mx` in design doc:
image
But we have pretty big `margin-x`:
image

T-Damer, line 71: Looks like the container grows when the content changes, that shouldn't happen:

https://user-images.githubusercontent.com/49658988/186421918-cfd5cc6d-7082-47e7-a252-bbb3c7c1d58e.mp4

T-Damer: Almost fixed it, container still grows a little, we have too many `w-full`, that's bad

src/components/Card.tsx
T-Damer: Okay, this one seems to be fixed

src/pages/Main.tsx
```diff
@@ -2,43 +2,56 @@ import Badges from 'components/badges/Badges'
...
-      <ProofsCard />
-      <CardSeparator
-        numberOfLines={3}
-        gradient="accent-to-secondary"
-        vertical={!lg}
-      />
-      <Badges />
-      {!lg && (
-        <>
+    <div>
```
T-Damer, line 34: I think we need to simplify this construction

src/components/Card.tsx
tuansydau: I finally fixed this, just added a maxWidth on the card container and let the cards take the full width

src/pages/Main.tsx
```diff
@@ -2,43 +2,60 @@ import Badges from 'components/badges/Badges'
...
+const cardsContainer = classnames(
   display('flex'),
   flexDirection('flex-col', 'lg:flex-row'),
   alignItems('items-center', 'lg:items-stretch'),
-  justifyContent('lg:justify-center')
+  justifyContent('lg:justify-center'),
+  margin('mx-10')
+)
+
+const mainContainer = classnames(width('w-full'), maxWidth('max-w-app-content'))
```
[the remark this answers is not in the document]
tuansydau: I had this happen a bit on large screen sizes, was able to fix it

src/components/Card.tsx
tuansydau: Unfortunately I haven't been able to figure out why this is - will this block launch? If not I can log it in a separate trello ticket

src/pages/Main.tsx
tuansydau: It says you marked this as resolved, was this on purpose? If not I'll work on simplifying this in the morning

src/components/Fade.tsx
```diff
@@ -18,7 +19,8 @@ const fade = (bottom = true) =>
     position('sticky'),
     inset(bottom ? '-bottom-2' : '-top-1', 'left-0', 'right-0'),
     height(bottom ? 'h-6' : 'h-4'),
-    zIndex('z-10')
+    zIndex('z-10'),
+    pointerEvents('pointer-events-none')
```
T-Damer, line 23: Click through fade is better UX as for me

src/components/Line.tsx
```diff
@@ -18,7 +14,7 @@ const line = (gradientDirection: 'to-left' | 'to-right', small?: boolean) => {
     backgroundImage(
       gradientDirection === 'to-left' ? 'bg-gradient-to-l' : 'bg-gradient-to-r'
     ),
-    display('hidden', 'xs:block')
+    displayFrom('xs')
```
T-Damer, line 17: Works fine, checked:
image
image

src/index.css
```diff
@@ -56,6 +59,14 @@ body {
     )
     0 0/3rem 3rem;
 }
+svg {
+  fill: none;
```
T-Damer, line 63: This can be overwritten by inline styles (or classes)

src/index.css
```diff
@@ -88,6 +99,10 @@ body {
     color: var(--tertiary);
     text-decoration: underline;
   }
+  .hover-menu-element:not(:disabled):hover {
```
T-Damer, line 102: https://user-images.githubusercontent.com/49658988/187470243-5970ab2e-525b-459f-9ec0-5283acb7acbe.mp4

T-Damer (commented):
https://user-images.githubusercontent.com/49658988/187626325-cad257d8-3dc6-4568-b61f-10dfa0d6570b.mp4

Still resizes if block is too small, we need to do something with it

src/classnames/tailwind.ts
```diff
@@ -7386,7 +7393,10 @@ export type TWidth =
   | 'w-card'
   | 'w-mobile-card'
   | 'w-full-125'
+  | 'w-full-105'
+  | 'w-screen-93'
   | 'w-screen-80'
+  | 'w-screen-45'
```
smyzin, line 7399: Should we stick to the screen sizes?

src/components/Card.tsx
```diff
@@ -89,15 +97,14 @@ const cardContainer = (
       'py-5': paddingType === 'normal',
       'py-8': typeof paddingType === 'undefined',
     }),
-    width(
-      thin ? 'sm:!w-thin-card' : 'sm:w-card',
-      thin ? 'xs:w-thin-mobile' : 'w-mobile-card',
-      { 'w-32': thin }
-    ),
+    useAppStyles ? undefined : thin ? thinWidthStyles : defaultWidthStyles,
```
smyzin, line 100: Can it be simplified and moved to the `const` variable? Like:
```
const widthStyle = () => {
  if (useAppStyles) return undefined
  return thin ? thinWidthStyles : defaultWidthStyles
}
```
I think it will be more convenient to use

src/components/proofs/Proof.tsx
```diff
@@ -27,10 +27,15 @@ import classnames, {
...
   width,
+  wordBreak,
 } from 'classnames/tailwind'
 import getEtherscanAddressUrl from 'helpers/getEtherscanAddressUrl'
 
-const proofName = classnames(display('flex'), flex('flex-1'))
+const proofName = classnames(
+  display('flex'),
+  flex('flex-1'),
+  wordBreak('break-words')
```
smyzin, line 37: You may use `hyphens` prop for the `ContractName` component. But this is an option too

smyzin (commented):
Found several UI bugs. Left the description and screenshots in the DM

src/classnames/tailwind.ts
T-Damer: It's an easy way to handle resizing without messing with `min` and `max` width

T-Damer: MinMax is hard to adapt on all screen sizes + have resizing that won't be triggered by inner changes (for example, expanding Email block when you have `min-max` + `width-full` will expand the block itself, but this shouldn't happen, this is why we should strictly limit the `width`, but keep it resizeable on different screens, which can be achieved only by using `vw`

src/components/proofs/Proof.tsx
T-Damer: Let's keep it like this for now

tailwind.config.js
```diff
@@ -41,10 +52,12 @@ module.exports = {
         11: '3.188rem',
       },
       screens: {
-        xxs: '17.5rem',
+        xxs: '17.6rem',
```
T-Damer, line 55: For some reason, it also touched the `fold` device, so I increased it a bit

tailwind.config.js
```diff
@@ -98,12 +111,16 @@ module.exports = {
         card: '31.75rem',
       },
       width: {
+        30: '7.5rem',
         'thin-mobile': '10.75rem',
-        'thin-card': '11.688rem',
+        'thin-card': '11.75rem',
```
T-Damer, line 116: This is actually a fine size of those thin cards on landing

smyzin (approved):

PR #441, opened 2022-08-17
fix generic gsn error

upacyxou:
- now "oops" message not showing, always displaying an actual error

### DIFF
--- a/src/helpers/parseGSNError.ts
+++ b/src/helpers/parseGSNError.ts
@@ -6,10 +6,13 @@ function transformRelayErrorMessage(message: string) {
     ?.join('\n')
 }
 
-export default function (gSNError: string) {
+function extractMetamaskReason(gSNError: string) {
   const metaMaskMessage = 'MetaMask Message Signature:'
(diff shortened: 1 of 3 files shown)


src/helpers/parseGSNError.ts
```diff
@@ -0,0 +1,15 @@
...
+  return message
+    .split('stack')
+    ?.filter((_, i) => i % 2 === 0)
+    ?.join('\n')
+}
+
+export default function (gSNError: string) {
+  const metaMaskMessage = 'MetaMask Message Signature:'
+  if (gSNError.includes(metaMaskMessage)) {
+    return gSNError.split(metaMaskMessage)[1]?.split(' stack:')[0]
```
T-Damer, line 12: Can we extract this into function with proper name to give better understanding on what is going on here?

T-Damer (approved):

src/helpers/parseGSNError.ts
upacyxou: sure

PR #445, opened 2022-08-18
fix gradient border in email button

upacyxou:
- gradient border returned to "submit email" button


T-Damer (approved):

src/components/EmailForm.tsx
```diff
@@ -30,19 +31,21 @@ export default function ({
...
-        fullWidth
-        center
-        small={submitType !== 'primary'}
-        type={submitType}
-        disabled={!emailIsValid}
-        onClick={() => onSubmit(email)}
-      >
-        {submitText}
-      </Button>
+      <GradientButton>
```
smyzin, line 34: Maybe we should name it `GradientBorder` instead?

src/components/GradientButton.tsx
```diff
@@ -10,7 +10,6 @@ import classnames, {
 } from 'classnames/tailwind'
 
 const parentButtonWrapper = classnames(
-  width('w-fit'),
```
smyzin, line 13: This style makes it fit the button on small screens. May you test that everything is fine, where that component is used?

upacyxou: sure, I tested it. The only two places where we're using it is CTA Buttons and "Submit emal" button, I tested them both and it's works fine on any resolution up to 280px width

src/components/EmailForm.tsx
upacyxou: I think you're right, it'll have more meaning. I'll rename it, ty

smyzin (approved):
Lgtm

PR #448, opened 2022-08-19
fix changing network behavior

upacyxou:
- doesn't prompt user to change network at page load, only when user clicking at "change network"
- refactored and reworked so changing network is more stable now


src/stores/WalletStore.ts
```diff
@@ -38,24 +36,39 @@ class WalletStore extends PersistableStore {
...
-      if (this.needNetworkChange)
-        throw new Error(
-          ErrorList.wrongNetwork(
-            (await provider.getNetwork()).name,
-            env.VITE_ETH_NETWORK
+      // We need this "any" networks so we could handle when users switch between networks with listeners without errors
+      provider = new Web3Provider(instance, 'any')
+
+      this.account = await this.getAccount()
+      if (!(await this.isNetworkRight())) {
```
T-Damer, line 60:
```TS
if (await this.isRightNetwork()) return
...
```
Less nesting = better readability (usually) + you don't need to convert boolean here, so  my code is a little faster

src/components/navbar/Wallet.tsx
```diff
@@ -79,7 +79,13 @@ const AccountContainer = ({ account }: { account?: string }) => {
   return (
     <div
       className={accountLinkContainer}
-      onClick={async () => await WalletStore.connect(true)}
+      onClick={async () => {
+        if (needNetworkChange) {
```
T-Damer, line 83: This is boilerplate code, we should but it into `WalletStore`

T-Damer (commented):
Let's clean up

src/stores/WalletStore.ts
```diff
@@ -38,24 +36,39 @@ class WalletStore extends PersistableStore {
     return web3Modal.cachedProvider
   }
 
+  private async getUserNetworkName() {
+    return (await provider.getNetwork()).name
+  }
+
+  private async isNetworkRight() {
```
T-Damer, line 43: Let's place `private` methods near each other, we have a couple of them down below

src/stores/WalletStore.ts
```diff
@@ -38,24 +36,39 @@ class WalletStore extends PersistableStore {
...
+      provider = new Web3Provider(instance, 'any')
+
+      this.account = await this.getAccount()
+      if (!(await this.isNetworkRight())) {
+        this.needNetworkChange = true
+        this.account = undefined
+        handleError(
+          new Error(
+            ErrorList.wrongNetwork(
+              await this.getUserNetworkName(),
```
T-Damer, line 66: Why won't you return the network name from this function, so we don't need to call it twice? Like if network is wrong, the function calller will have userNetwork and our network, and we'll pass them here

src/stores/WalletStore.ts
```diff
@@ -158,13 +155,27 @@ class WalletStore extends PersistableStore {
       void this.handleAccountChanged()
     })
     provider.on('disconnect', (error: unknown) => {
+      // Sometimes this error fires when user switching between networks too fast, we should not clean data in this case
+      if (
+        error instanceof Error &&
```
T-Damer, line 160: I think we can extract his into `helpers`, maybe we have something like this in `handleErro`

src/stores/WalletStore.ts
```diff
@@ -38,24 +36,39 @@ class WalletStore extends PersistableStore {
...
+    return (await provider.getNetwork()).name
+  }
+
+  private async isNetworkRight() {
+    const network = env.VITE_ETH_NETWORK
+    const userNetwork = await this.getUserNetworkName()
+    return network === userNetwork
+  }
+
   async connect(clearCachedProvider = false) {
```
T-Damer, line 49: Btw, looks like we use this as `true` usually, please check it and chagne if it is

src/components/navbar/Wallet.tsx
T-Damer: This will look like:
```TS
WalletStore.connect({clearCachedProvider: true, needNetwrokChange })
```

upacyxou: I agree that it's a boilerplate and we need to extract it to WalletStore, but I don't think that it's a good idea to put it in `connect` function. The only one thing we need to do is to switch the network, not to connect anything (provider or whatever). I'll create a function like `connectOrChangeNetwork`.

src/stores/WalletStore.ts
upacyxou: Will fix this, ty

src/stores/WalletStore.ts
upacyxou: But we don't want to return, we want to subscribe to our provider anyway, even if the network is not right

src/stores/WalletStore.ts
upacyxou: Imo, returning network name from `isNetworkRight` function isn't a good idea, because it's not self-explanatory. It's better for reading and understanding to use separate calls for getting a network name imo. One function do only one thing. Either returning the network name, either returning is network a right one. But not two at the same time.

src/stores/WalletStore.ts
upacyxou: Nice idea, ty

src/stores/WalletStore.ts
upacyxou: Will check, ty

src/stores/WalletStore.ts
T-Damer: Well, you're right, thanks :)

src/stores/WalletStore.ts
T-Damer: I can see, then, extract the `await this.isRightNetwork()` into `const`, please, this will increase readability

src/components/navbar/Wallet.tsx
T-Damer: Okay, thanks :)

src/stores/WalletStore.ts
```diff
@@ -158,13 +155,27 @@ class WalletStore extends PersistableStore {
       void this.handleAccountChanged()
     })
     provider.on('disconnect', (error: unknown) => {
+      // Sometimes this error fires when user switching between networks too fast, we should not clean data in this case
+      if (
+        error instanceof Error &&
+        error.message.includes('Attempting to connect.')
+      ) {
+        handleError(error)
+        return
```
T-Damer, line 164: Btw, you can do `return handleError(error)` as a shortcut, I think it won't create any side effects

src/stores/WalletStore.ts
```diff
@@ -158,13 +155,27 @@ class WalletStore extends PersistableStore {
...
     })
     provider.on('chainChanged', async () => {
+      if (await this.isNetworkRight()) {
+        this.needNetworkChange = false
+        this.account = await this.getAccount()
+        return
+      }
       this.account = undefined
-      await this.connect()
+      this.needNetworkChange = true
```
T-Damer, line 178: Looks like you're doing same thing in `.connect` method (I mean cleaning the `localstorage` data), let's extract it into function like `private resetData`

src/stores/WalletStore.ts
```diff
@@ -110,40 +123,24 @@ class WalletStore extends PersistableStore {
...
+    this.account = await this.getAccount()
     NotificationsStore.showTwitterShare = false
     this.walletLoading = false
   }
 
+  private async getAccount() {
+    return (await provider.listAccounts())[0]
+  }
+
+  async switchNetwork() {
```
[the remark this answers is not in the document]
upacyxou: You're right, nice catch

src/stores/WalletStore.ts
upacyxou: Got it. Sure, nice catch, ty

src/stores/WalletStore.ts
upacyxou: I guess you're right, ty

src/stores/WalletStore.ts
upacyxou: Actually, I don't think that it's a good idea to clear provider when it's not strictly defined, so I'll not change it to true as default

src/stores/WalletStore.ts
upacyxou: wdym by "cleaning localstorage data"? I'm changing only "account" and "needNetworkChange" value

src/components/navbar/Wallet.tsx
```diff
@@ -79,7 +79,9 @@ const AccountContainer = ({ account }: { account?: string }) => {
   return (
     <div
       className={accountLinkContainer}
-      onClick={async () => await WalletStore.connect(true)}
+      onClick={async () => {
+        await WalletStore.connectOrChangeNetwork(true, needNetworkChange)
```
T-Damer, line 83: Can we make the arguments as object, so we can understand how we connect the wallet without going into the store to check what these arguments mean?

src/helpers/isSwitchingNetworkError.ts
```diff
@@ -0,0 +1,9 @@
+export default function (error: unknown) {
+  // Sometimes this error fires when user switching between networks too fast, we should not clean data in this case
+  if (
+    error instanceof Error &&
+    error.message.includes('Attempting to connect.')
```
T-Damer, line 5: Hmm, what if some `axios` request will have such a message? Maybe we should change it?

src/stores/WalletStore.ts
```diff
@@ -38,24 +37,95 @@ class WalletStore extends PersistableStore {
     return web3Modal.cachedProvider
   }
 
-  async connect(clearCachedProvider = false) {
+  private async getUserNetworkName() {
```
T-Damer, line 40: Provider may be `undefined` (types say that it can't, but actually it's not initialized before `.connect` happens), if you try to call this method without knowing this moment, you'll fail. Let's pass the `provider` here as non-optional argument

src/helpers/isSwitchingNetworkError.ts
upacyxou: wdym? can you elaborate, please?

src/stores/WalletStore.ts
```diff
@@ -66,6 +136,14 @@ class WalletStore extends PersistableStore {
     }
   }
 
+  connectOrChangeNetwork(
```
MixailE, line 139: It seems that by calling `connectOrChangeNetwork` without any parameters, the network should change if necessary by default

upacyxou: oh, that's sound logical, I'll change it to `changeNetworkOrConnect`

MixailE: maybe it's better to replace `clearCachedProvider = false` with `useCachedProvider = true`?

src/stores/WalletStore.ts
```diff
@@ -38,24 +37,95 @@ class WalletStore extends PersistableStore {
...
-            env.VITE_ETH_NETWORK
+      // We need this "any" networks so we could handle when users switch between networks with listeners without errors
+      provider = new Web3Provider(instance, 'any')
+
+      this.account = await this.getAccount()
+      const isNetworkNotRight = !(await this.isNetworkRight())
+      if (isNetworkNotRight) {
+        this.needNetworkChange = true
+        this.account = undefined
+        handleError(
```
T-Damer, line 120: Seems like we can simplify it:
1) Let's make `env.VITE_ETH_NETWORK` be the default argument inside the `.wrongNetwork`
2) Maybe you can make `ErrorList` a function that will spawn `new Error`? Like this: `ErrorList({newError: true}).wrongNetwork(...)`

src/stores/WalletStore.ts
```diff
@@ -38,24 +37,95 @@ class WalletStore extends PersistableStore {
...
-          ErrorList.wrongNetwork(
-            (await provider.getNetwork()).name,
-            env.VITE_ETH_NETWORK
+      // We need this "any" networks so we could handle when users switch between networks with listeners without errors
+      provider = new Web3Provider(instance, 'any')
+
+      this.account = await this.getAccount()
+      const isNetworkNotRight = !(await this.isNetworkRight())
+      if (isNetworkNotRight) {
+        this.needNetworkChange = true
```
T-Damer, line 118: You have:
```TS
      this.account = undefined
      this.needNetworkChange = true
```
in `subscribeProvide ... provider.on('chainChanged')`, let's extract these two lines into `private` function

src/stores/WalletStore.ts
```diff
@@ -38,24 +37,95 @@ class WalletStore extends PersistableStore {
...
-      if (this.needNetworkChange)
-        throw new Error(
-          ErrorList.wrongNetwork(
-            (await provider.getNetwork()).name,
-            env.VITE_ETH_NETWORK
+      // We need this "any" networks so we could handle when users switch between networks with listeners without errors
+      provider = new Web3Provider(instance, 'any')
+
+      this.account = await this.getAccount()
+      const isNetworkNotRight = !(await this.isNetworkRight())
```
T-Damer, line 116: Seems hard to read, let's not extract this into `const`, use you did before:
```TS
if (!(await this.isNetworkRight())
```

src/stores/WalletStore.ts
```diff
@@ -38,24 +37,95 @@ class WalletStore extends PersistableStore {
...
+      handleError(error)
+    })
+
+    provider.on('accountsChanged', (accounts: string[]) => {
+      if (!accounts.length) this.clearData()
+
+      this.account = undefined
+      void this.handleAccountChanged()
+    })
+    provider.on('disconnect', (error: unknown) => {
```
T-Damer, line 87: btw, are you sure that `error` here is `unknown`, there is a `provider.on(error: Error)`, error has `Error` type, maybe here too?

T-Damer (commented):
So close, let's do this 🚀

src/stores/WalletStore.ts
upacyxou: ty, fixed

src/stores/WalletStore.ts
upacyxou: you're right, fixed

src/stores/WalletStore.ts
upacyxou: 1. Changed
2. I don't think that it's a good idea, it'll add a boilerplate for our entire codebase, while it's actully needed only here, in this place

src/helpers/isSwitchingNetworkError.ts
upacyxou: nvm, I just removed this file

src/stores/WalletStore.ts
upacyxou: I renamed, but I dont think that we need to rename it

src/stores/WalletStore.ts
```diff
@@ -38,24 +36,99 @@ class WalletStore extends PersistableStore {
     return web3Modal.cachedProvider
   }
 
-  async connect(clearCachedProvider = false) {
+  private async getUserNetworkName(provider: Web3Provider) {
+    return (await provider.getNetwork()).name
+  }
+
+  private async isNetworkRight(provider: Web3Provider) {
+    const network = env.VITE_ETH_NETWORK
```
T-Damer, line 44: Maybe we should create `const preferedNetwork = env.VITE_ETH_NETOWRK` above this `class`, so we don't need to use `env` every time?

src/stores/WalletStore.ts
```diff
@@ -66,6 +139,17 @@ class WalletStore extends PersistableStore {
     }
   }
 
+  changeNetworkOrConnect({
+    clearCachedProvider,
+    needNetworkChange,
+  }: {
+    clearCachedProvider: boolean
```
T-Damer, line 146: btw, you can set defaults for this

T-Damer (approved):

src/stores/WalletStore.ts
upacyxou: sure, nice idea. ty

PR #452, opened 2022-08-22
refactor stores

MixailE:


src/components/ContractName.tsx
```diff
@@ -1,17 +1,15 @@
...
   display,
   flexWrap,
   minWidth,
   wordBreak,
 } from 'classnames/tailwind'
-import getContractName from 'helpers/getContractName'
-import prettifyContractName from 'helpers/prettifyContractName'
-import truncateMiddleIfNeeded from 'helpers/truncateMiddleIfNeeded'
+import getContractName from 'helpers/network/getContractName'
+import prettifyContractName from 'helpers/network/prettifyContractName'
```
T-Damer, line 11: Smells like we don't need those two separate helpers, maybe we can chain them like this:
```TS
const contractName = getContractName({prettify: true, truncate: truncateOption})
```
?

src/components/badges/BadgesOwnedForContract.tsx
```diff
@@ -1,24 +1,24 @@
...
+import useContractTokens from 'hooks/useContractTokens'
 
 function BadgesOwnedForContractSuspended({
   contractAddress,
 }: {
   contractAddress: string
 }) {
-  const { addressToTokenIds } = useSnapshot(GoerliContractsStore)
+  const ownedIds = useContractTokens(contractAddress, BadgesContractsStore)
   const { account } = useSnapshot(WalletStore)
```
T-Damer, line 15: Maybe we need a blank line between hooks and logic?

src/components/badges/List.tsx
```diff
@@ -67,33 +57,16 @@ function BadgeListSuspended() {
...
-          <BadgeSection
-            title="Email derivatives"
-            minted={ownedEmailDerivativeContracts}
-            proofs={proofsAvailableToMint.filter(
-              (proof) => proof instanceof EmailProof
-            )}
-            onMinted={onMinted}
-          />
+          {Object.entries(ownedContractsByKey).map(([key, contracts]) => (
+            <>
```
T-Damer, line 61: I think it's better not to wrap `each` `BadgeSection`, but to wrap all `BadgeSection`s into `<></>`

T-Damer: So move this line to `L60`

src/components/badges/List.tsx
```diff
@@ -67,33 +57,16 @@ function BadgeListSuspended() {
...
+              <BadgeSection
+                title={data[key as keyof typeof data].title}
```
T-Damer, line 63: You're doing `as keyof typeof data` four times in your PR, let's place this into `data` as type:
`DataContractNames`, for example, or `DataKeys`

T-Damer: And let's use this type across the PR

src/components/badges/ShareToTwitterIfNeeded.tsx
```diff
@@ -32,7 +32,6 @@ const wideBlock = classnames(
 
 function ShareToTwitterIfNeededSuespended() {
   const { showTwitterShare } = useSnapshot(NotificationsStore)
-
```
T-Damer, line 35: Why don't you want a blank line between hooks and logic?)

src/components/owned/OwnerInfo.tsx
```diff
@@ -72,39 +70,32 @@ function BadgeTitle({ badge }: { badge: BaseBadgeContract }) {
 }
 
 function BadgeContent({ badge }: { badge: BaseBadgeContract }) {
-  if (badge instanceof EmailBadgeContract) {
+  if (badge.type === 'Email') {
```
T-Damer, line 73: Comparing to `string` is not 100% safe, maybe we can use `enums` for this? Just a suggestion, if you don't want to do it, you shoudn't :)

src/components/owned/OwnerInfo.tsx
```diff
@@ -72,39 +70,32 @@ function BadgeTitle({ badge }: { badge: BaseBadgeContract }) {
...
+  return (
+    <BodyText>
+      This is a zkNFT derivative. It means this person has been verified to own
+      at least one ‘
+      <ExternalLink url={getEtherscanAddressUrl(badge.original, network)}>
+        <AccentText color="text-secondary">
+          <ContractName address={badge.original} network={network} />
+        </AccentText>
+      </ExternalLink>
+      ‘ {network[0].toUpperCase() + network.slice(1)} NFT.
```
T-Damer, line 95: Uppercase first letter, let's do a `helper` for this, we may need it in the future

src/components/proofs/ReadyERC721ProofsList.tsx
```diff
@@ -1,25 +1,22 @@
...
+  const { proofsCompleted } = useSnapshot(ProofStore)
 
-  const ERC721ProofsCompleted = networkPick(
-    network,
-    goerliERC721ProofsCompleted,
-    mainnetERC721ProofsCompleted
-  )
+  const filtered = proofsCompleted.filter(
+    (proof) => proof instanceof ERC721Proof && proof.network === network
+  ) as ERC721Proof[]
```
T-Damer, line 13: Hm, do you need to convert it? You already checked for `instanceof` 🤔

T-Damer: And let's name this `const` properly, I know that you filtered something, but what is the result of this filtering? I can read the code to understand, but isn't it's better to just read the name of the variable and understand it faster?

src/data.ts
```diff
@@ -0,0 +1,34 @@
...
+export enum BadgeSourceType {
+  Email = 'Email',
+  ERC721 = 'ERC721',
+}
+
+export default {
+  ExternalERC721: {
+    ledger: env.VITE_SC_EXTERNAL_ERC721_LEDGER_CONTRACT_ADDRESS,
+    title: 'Mainnet NFT derivatives',
+    source: {
```
T-Damer, line 13: Why do you need to nest `type` and `network` in `source`? If there's no reason, we shouldn't nest it

T-Damer: As I can see, there's no reason, so let's remove it and use directly from the object

src/components/badges/ShareToTwitterIfNeeded.tsx
T-Damer: We have them at most places as I know, so let's get it back :)

src/components/badges/List.tsx
```diff
@@ -21,32 +20,23 @@ import walletStore from 'stores/WalletStore'
...
   )
 
   const proofsAvailableToMint = useProofsAvailableToMint()
 
   const hasProofs = proofsAvailableToMint.length
-  const hasDerivatives =
-    ownedExternalERC721DerivativeContracts.length +
-    ownedEmailDerivativeContracts.length +
-    ownedERC721DerivativeContracts.length
+  const hasDerivatives = Object.values(ownedContractsByKey).reduce(
```
T-Damer, line 36: Maybe we can have a function to explain what is going on here?

MixailE: I simplified a little

src/components/ContractName.tsx
```diff
@@ -41,21 +38,13 @@ function ContractNameSuspended({
   clearType,
   network,
 }: ContractNameProps) {
-  const { emailDerivativeContracts = [], ERC721derivativeContracts = [] } =
-    useSnapshot(SealCredStore)
   // We will always get a string
-  let contractName = getContractName(address, network) as string
+  let contractName = getContractName(address, network, truncate) as string
```
T-Damer, line 42: Why do we need to convert `contractName as string` here? As I can see, the call stack doesn't return `undefined`, so we don't need this?

src/components/ContractName.tsx
```diff
@@ -41,21 +38,13 @@ function ContractNameSuspended({
...
 
   if (clearType) {
-    if (contractName && emailDerivativeContracts.includes(address))
-      contractName = contractName.replace(' email', '')
-
-    if (contractName && ERC721derivativeContracts.includes(address))
-      contractName = contractName.replace(' (derivative)', '')
+    contractName = contractName.replace(/ (email|\(derivative\))$/, '')
```
T-Damer, line 45: I guess now we can get rid of `{}`, but it's up to you :)

src/components/ENSAddress.tsx
```diff
@@ -15,9 +15,10 @@ function ENSAddressSuspended({
   truncate,
   truncateSize,
 }: ENSAddressProps & { truncate?: boolean; truncateSize: number }) {
-  const { eNSNames } = useSnapshot(GoerliENSStore)
+  const { eNSNames } = useSnapshot(ENSStore.networks[Network.Goerli])
```
T-Damer, line 18: Maybe let's use `{ensNames}`? It's pretty strange naming here :)

src/components/badges/BadgesOwnedForContract.tsx
```diff
@@ -1,24 +1,26 @@
...
+
   if (!account) {
     return <BadgesOwnedForContractLoading contractAddress={contractAddress} />
   }
-  if (!addressToTokenIds) {
+  if (ownedIds.length === 0) {
     return <BadgesOwnedForContractLoading contractAddress={contractAddress} />
```
T-Damer, line 21: Seems like the return statement here is the same, as on `L18`, let's combine `if` info:
```TS
if (!account || ownedIds.length === 0) 
  return <BadgesOwnedForContractLoading contractAddress={contractAddress} />
```

src/components/badges/ListTitle.tsx
```diff
@@ -8,15 +8,23 @@ import useContractsOwned from 'hooks/useContractsOwned'
...
-  const { derivativeContracts = [] } = useSnapshot(SealCredStore)
+  const { derivativeContracts } = useSnapshot(SealCredStore)
   const proofsAvailableToMint = useProofsAvailableToMint()
-  const contractsOwned = useContractsOwned(GoerliContractsStore)
+  const contractsOwned = useContractsOwned(BadgesContractsStore)
 
-  const ownedDerivativeContracts = derivativeContracts.filter(
-    (contractAddress) => contractsOwned.includes(contractAddress)
+  const ownedDerivativeContracts = Object.values(derivativeContracts).reduce(
+    (chain, contracts) =>
```
T-Damer, line 16: `chain` may be the verb, like chain operations, can we call it `chainName`?

T-Damer: Or something more explanatory

src/components/badges/ListTitle.tsx
```diff
@@ -8,15 +8,23 @@ import useContractsOwned from 'hooks/useContractsOwned'
...
+      chain.concat(
+        contracts.filter((contractAddress) =>
+          contractsOwned.includes(contractAddress)
+        )
+      ),
+    []
   )
 
-  const hasUnminted = proofsAvailableToMint.length > 0
+  const hasUnminted = Object.values(proofsAvailableToMint).some(
```
T-Damer, line 25: You're doing this operation in `badges/List.tsx/L35`, maybe it's wors extracting this operation into `helpers`? Something like `checkObjectValuesHasSome`

T-Damer: If it's not easy to do, let's leave it like this

src/components/owned/OwnerInfo.tsx
```diff
@@ -29,82 +28,67 @@ const walletBox = classnames(
...
-      </HeaderText>
-    </>
+function BadgeTitle({
+  badge: { type, derivative },
+}: {
+  badge: BaseBadgeContract
+}) {
+  const { ownerTitle } = data[type]
+
+  if (!ownerTitle) return null
```
T-Damer, line 52: Will destructuring work if there's nothing to destructure?

src/components/owned/OwnerInfo.tsx
```diff
@@ -29,82 +28,67 @@ const walletBox = classnames(
...
-              <ContractName
-                hyphens
-                address={badge.address}
-                network={Network.Goerli}
-              />
-            </AccentText>
-          </ExternalLink>
-        </HeaderText>
-      </>
+function replace(
```
T-Damer, line 31: Maybe let's give it a more explanatory name? Something like `replaceKeywordsInTemplateJSX`?

src/components/owned/OwnerInfo.tsx
```diff
@@ -29,82 +28,67 @@ const walletBox = classnames(
...
+  template: string,
+  keyword: string,
+  element: JSXInternal.Element
```
T-Damer, line 34: As I know, you can import `Element` from `preact/compat`, in SCE I can use it as a global declaration without importing

src/components/proofs/ProofList.tsx
```diff
@@ -19,11 +19,12 @@ export function ProofListSuspended() {
...
         <div className={space('space-y-2')}>
-          {account && (
-            <>
-              <ERC721ProofSection account={account} network={Network.Mainnet} />
-              <ERC721ProofSection account={account} network={Network.Goerli} />
-            </>
+          {Object.values(data).map(
+            ({ badgeType, network }) =>
+              badgeType === BadgeSourceType.ERC721 &&
+              account && (
```
T-Damer, line 25: I think we should check the account before the mapping, so we don't try to map without connected account

src/components/proofs/ReadyERC721ProofsList.tsx
```diff
@@ -1,25 +1,22 @@
...
-    mainnetERC721ProofsCompleted
-  )
+  const allERC721ProofsByNetwork = proofsCompleted.filter(
+    (proof) => proof instanceof ERC721Proof && proof.network === network
+  ) as ERC721Proof[]
 
   return (
     <>
-      {!!ERC721ProofsCompleted?.length && (
+      {!!allERC721ProofsByNetwork?.length && (
```
T-Damer, line 17: Why we need to do `!!` here? If the `length` is `0` or `allERC721ProofsByNetwork is undefined`, then it won't return anything, isn't it?

src/data.ts
```diff
@@ -0,0 +1,38 @@
...
+    badgeType: BadgeSourceType.ERC721,
+    network: Network.Mainnet,
+  },
+  Email: {
+    ledger: env.VITE_SC_EMAIL_LEDGER_CONTRACT_ADDRESS,
+    title: 'Email derivatives',
+    ownerTitle: 'This wallet belongs to someone with {derivative}',
+    ownerContent:
+      'This is a zkNFT derivative of an email. It means this person has been verified own a ‘{original}‘ email.',
+
```
T-Damer, line 25: I think we don't need this blank line :)

src/helpers/contracts/dataShapeObject.ts
```diff
@@ -0,0 +1,20 @@
+import { DataKeys } from 'models/DataKeys'
+import data from 'data'
+
+type ValueOf<T> = T[keyof T]
+
+export default function <V>(
```
T-Damer, line 6: Let's add a description for the function or let's rename it :)

src/helpers/contracts/ledgerContracts.ts
```diff
@@ -0,0 +1,18 @@
...
+  Ledger__factory,
+} from '@big-whale-labs/seal-cred-ledger-contract'
+import Network from '@big-whale-labs/stores/dist/models/Network'
+import data from 'data'
+import networks from 'networks'
+
+export default Object.entries(data).reduce(
+  (prev, [name, { ledger }]) => ({
+    ...prev,
+    [name]: Ledger__factory.connect(
```
T-Damer, line 12: Why don't we include this into our `data`?

T-Damer: Something like:
```TS
ERC721: {
  ledger: env.SC_ERC721_LEDGER,
  contract: Ledger__factory.connect(env.SC_ERC721_LEDGER, networks[Network.Goerli].defaultProvider
  }
```

src/networks.ts
```diff
@@ -0,0 +1,21 @@
+import { ETH_RPC, ETH_RPC_MAINNET } from '@big-whale-labs/constants'
+import Network from 'models/Network'
+import createProvider from 'helpers/providers/createProvider'
+import env from 'helpers/env'
+
+export default {
+  [Network.Mainnet]: {
+    network: Network.Mainnet,
+    ...createProvider(Network.Mainnet, {
+      default: ETH_RPC_MAINNET,
```
T-Damer, line 10: I think we shouldn't use `@bwl/constants` here, use `env.ETH_RPC_MAINNET`, it will have `@bwl/contants` + your own `.env` if you need it

src/hooks/useProofAddressesAvailableToCreate.ts
```diff
@@ -1,47 +1,44 @@
...
 
 export default function (network?: Network) {
-  const { goerliERC721ProofsCompleted, mainnetERC721ProofsCompleted } =
-    useSnapshot(ProofStore)
+  const { eRC721ProofsCompleted } = useSnapshot(ProofStore)
+  const { derivativeContracts = [] } = useSnapshot(SealCredStore)
+
   let contractsOwned: readonly string[]
   switch (network) {
     case Network.Mainnet:
```
T-Damer, line 15: If `Mainnet` has empty case, maybe we should remove this line at all?

src/hooks/useProofsAvailableToMint.ts
```diff
@@ -1,37 +1,24 @@
...
+  return dataShapeObject((ledgerKey) =>
+    proofsCompleted.filter((proof) => {
+      const proofLedgerName = proof.dataType
+      if (proofLedgerName !== ledgerKey) return false
+      const ledger = ledgers[proofLedgerName]
+      const record = ledger[proof.origin]
       return (
-        !emailLedger[proof.domain] ||
-        !contractsOwned.includes(emailLedger[proof.domain].derivativeContract)
+        !record || !contractsOwned.includes(record.derivative.toLowerCase())
```
T-Damer, line 20: Can you make a comment here, why we need to do `!`?

T-Damer: And maybe we should convert all the `0x...` addresses `toLowerCase` everywhere?

src/models/SCLedger.ts
```diff
@@ -0,0 +1,6 @@
+export default interface Ledger {
+  [original: string]: {
+    original: string
```
T-Damer, line 3: We were able to get rid of it in `SCE`, maybe we can do it here too?

src/stores/ContractsStore.ts
```diff
@@ -1,48 +1,44 @@
...
+import transformObjectValues from 'helpers/transformObjectValues'
 
-export const GoerliContractsStore = proxy(
-  new ContractsStore(goerliDefaultProvider, goerliHeavyProvider, Network.Goerli)
-).makePersistent(env.VITE_ENCRYPT_KEY)
+const proxyNetworks = transformObjectValues(
+  networks,
+  ({ network, heavyProvider, defaultProvider }) =>
+    proxy(
+      new ContractsStore(defaultProvider, heavyProvider, network)
```
T-Damer, line 14: Hm, maybe we can put default data inside `ContractStore`? If it's hard to do, let's leave as it is

src/stores/ContractsStore.ts
```diff
@@ -1,48 +1,44 @@
...
-mainnetDefaultProvider.on('block', async (blockNumber: number) => {
-  MainnetContractsStore.currentBlock = blockNumber
-  if (WalletStore.account) {
-    await MainnetContractsStore.fetchMoreContractsOwned(WalletStore.account)
-  }
-})
+for (const { defaultProvider, network } of Object.values(networks)) {
+  defaultProvider.on('block', async (blockNumber: number) => {
+    ContractsNetworkStore.networks[network].currentBlock = blockNumber
+    if (WalletStore.account) {
```
T-Damer, line 36: Let's remove `{}` from here, let's be consistent about `{}`, if we don't use them mostly, let's not do it

src/stores/ProofStore.ts
```diff
@@ -57,25 +57,25 @@ class ProofStore extends PersistableStore {
     return selected
   }
 
-  get ERC721ProofsCompleted() {
+  get eRC721ProofsCompleted() {
```
T-Damer, line 60: Let's name it `erc721ProofsCompleted`

src/stores/ProofStore.ts
```diff
@@ -57,25 +57,25 @@ class ProofStore extends PersistableStore {
...
     const selected = []
     for (const proof of this.proofsCompleted) {
       if (proof instanceof ERC721Proof) selected.push(proof)
     }
     return selected
   }
 
-  get goerliERC721ProofsCompleted() {
+  get goerlieRC721ProofsCompleted() {
```
T-Damer, line 68: Maybe `goerliErc721ProofsCompleted`? Let's keep the `camelCase`

T-Damer (commented):
There's always a room for improvement, it looks cool already, let's make it even better 🚀

src/hooks/useAllContractsOwned.ts
```diff
@@ -0,0 +1,15 @@
+import { useSnapshot } from 'valtio'
+import ContractsNetworkStore from 'stores/ContractsStore'
+
+export default function () {
+  const { networks } = useSnapshot(ContractsNetworkStore)
+
+  const addressToTokenIds = Object.values(networks).reduce(
+    (res, network) => ({ ...res, ...network.addressToTokenIds }),
```
T-Damer, line 8: Btw, what is `res`? Maybe we can have another name for `res`, isn't this is a `prev`?

src/components/ENSAddress.tsx
MixailE: All variables start with a small letter. This `eNSNames` property is exported from the library class

src/hooks/useProofAddressesAvailableToCreate.ts
MixailE: Then the case won't work for the mainnet, right?)

src/models/SCLedger.ts
MixailE: Are yor sure?) https://github.com/BigWhaleLabs/seal-cred-echo-frontend/blob/main/src/models/Ledger.ts

src/stores/ContractsStore.ts
MixailE: This class `ContractStore` is imported from the library

src/hooks/useProofsAvailableToMint.ts
MixailE: I'm not sure what should be convert everywhere, this should be done when we receive/store data from somewhere

src/networks.ts
MixailE: Most likely we need to make a separate variable for these purposes

src/components/owned/OwnerInfo.tsx
MixailE: Types are supposed to tell if it's not possible, right?

src/components/ContractName.tsx
```diff
@@ -41,20 +38,10 @@ function ContractNameSuspended({
...
-  if (clearType) {
-    if (contractName && emailDerivativeContracts.includes(address))
-      contractName = contractName.replace(' email', '')
-
-    if (contractName && ERC721derivativeContracts.includes(address))
-      contractName = contractName.replace(' (derivative)', '')
-  }
-
-  contractName = prettifyContractName(contractName, truncate)
+  if (clearType)
```
T-Damer, line 43: Actually, maybe we should use `clearType` directly inside `getContractName` function?

src/helpers/network/getContractName.ts
```diff
@@ -0,0 +1,18 @@
+import { useSnapshot } from 'valtio'
+import ContractMetadataStore from 'stores/ContractMetadataStore'
+import Network from 'models/Network'
+import networks from 'networks'
+import prettifyContractName from 'helpers/network/prettifyContractName'
+
+export default function (address: string, network: Network, truncate = false) {
```
T-Damer, line 7: Maybe let's make an object argument? Because you need to check the order each time you want do use the helper

src/helpers/contracts/dataShapeObject.ts
```diff
@@ -0,0 +1,20 @@
+import { DataKeys } from 'models/DataKeys'
+import data from 'data'
+
+type ValueOf<T> = T[keyof T]
+
+export default function dataShapeObject<V>(
```
T-Damer, line 6: This helper looks similar to `ledgerContracts.ts`, maybe we can use only one of them? If it's too complicated to do/to use, let's leave as it is

src/helpers/providers/createProvider.ts
```diff
@@ -0,0 +1,20 @@
...
+  networkToRPC: { [network in 'heavy' | 'default']: string }
+) {
+  return {
+    network,
+    defaultProvider: new providers.JsonRpcProvider(
+      networkToRPC.default,
+      network.toLowerCase()
+    ),
+    heavyProvider: alchemyOrJsonProvider(
+      network.toLowerCase(),
```
T-Damer, line 16: Let's not mix the order of the arguments:
1st) networkToRpc
2nd) network

src/hooks/useBadgeContracts.ts
```diff
@@ -1,42 +1,38 @@
...
-      ...Object.values(ERC721Ledger),
-      ...Object.values(externalERC721Ledger),
-    ].reduce(
-      (result, { derivativeContract, originalContract }) => ({
-        ...result,
-        [derivativeContract]: new ERC721BadgeContract(
-          derivativeContract,
-          originalContract,
-          ERC721Ledger[originalContract] ? Network.Goerli : Network.Mainnet
+    return Object.entries(ledgers).reduce(
```
T-Damer, line 16: Basically we're trying to fit to a type we've used before, maybe we can change `data`/`ledger`, so we don't need to change their types at all?

src/components/proofs/ERC721ProofSection.tsx
```diff
@@ -16,11 +16,11 @@ export function ERC721ProofSection({
   account: string
   network: Network
 }) {
-  const { ERC721ProofsCompleted } = useSnapshot(ProofStore)
+  const { eRC721ProofsCompleted } = useSnapshot(ProofStore)
```
T-Damer, line 19: Well, names still look awkward, we need to do something with them

src/models/SCLedger.ts
```diff
@@ -0,0 +1,6 @@
+export default interface Ledger {
```
T-Damer, line 1: btw, why do we need to store is like `[key]: value`?
Why don't we store everything like:
```TS
export default interface Ledger {
  original: string,
  derivative: string,
}
```
We can use it easily + will shortcut the `Object.keys Object.values` operations, but in this case we might lose easy access by `Ledger[originalAddress]`, so we need to think about it

src/components/owned/OwnerInfo.tsx
```diff
@@ -29,82 +28,73 @@ const walletBox = classnames(
...
+    </ExternalLink>
+  )
+
+  const title = replaceKeywordsInString(
+    template,
+    '{derivative}',
+    derivativeLink
   )
+
+  return <HeaderText extraLeading>{title}</HeaderText>
```
smyzin, line 68: It looks a little bit complex. May we do smt with it?

src/components/owned/OwnerInfo.tsx
```diff
@@ -29,82 +28,73 @@ const walletBox = classnames(
...
-        </ExternalLink>
-        ‘ {badge.network[0].toUpperCase() + badge.network.slice(1)} NFT.
-        <ShareCTAButtons />
-      </BodyText>
-    )
-  }
+  const originalLink = (
+    <ExternalLink url={getEtherscanAddressUrl(original, network)}>
+      <AccentText color="text-secondary">
+        <ContractName address={original} network={network} />
```
smyzin, line 85: Don't we need a `hyphens` here? As for component blow

src/components/owned/OwnerInfo.tsx
```diff
@@ -29,82 +28,73 @@ const walletBox = classnames(
...
-              network={badge.network}
-            />
-          </AccentText>
-        </ExternalLink>
-        ‘ {badge.network[0].toUpperCase() + badge.network.slice(1)} NFT.
-        <ShareCTAButtons />
-      </BodyText>
-    )
-  }
+  const originalLink = (
```
smyzin, line 82: This component looks identically like `**derivativeLink**`. Could we do smt with it?

src/components/proofs/ERC721ProofSection.tsx
```diff
@@ -1,73 +1,55 @@
...
 
-export function ERC721ProofSection({
-  account,
-  network,
-}: {
-  account: string
-  network: Network
-}) {
-  const { ERC721ProofsCompleted } = useSnapshot(ProofStore)
+export function ERC721ProofSection({ dataKey }: { dataKey: DataKey }) {
```
smyzin, line 13: Let's rename it to `ERC721ProofSectionSuspended`

src/components/proofs/ERC721ProofSection.tsx
```diff
@@ -1,73 +1,55 @@
...
+      proofs={proofsCompleted as Proof[]}
+      originals={networkProofAddressesAvailableToCreate}
+      onCreate={onCreate}
+      dataKey={dataKey}
+      nothingToGenerateText="No NFTs to proof"
+    />
   )
 }
 
 export default function ERC721ProofSectionSuspended({
```
smyzin, line 34: And this one you may leave with out the name

src/helpers/network/prettifyContractName.ts
```diff
@@ -1,5 +1,5 @@
 import { utils } from 'ethers'
-import truncateMiddleIfNeeded from 'helpers/truncateMiddleIfNeeded'
+import truncateMiddleIfNeeded from 'helpers/network/truncateMiddleIfNeeded'
```
smyzin, line 2: Why did you place it into `network`?

src/components/badges/BadgeBlock.tsx
```diff
@@ -1,70 +1,54 @@
...
+import data from 'data'
 import handleError from 'helpers/handleError'
 
 function Badge({
   proof,
   onMinted,
   onMintFailed,
 }: {
-  proof: BaseProof
+  proof: Proof
```
T-Damer, line 20: Extract it into `inerface`, you have same thing [here](https://github.com/BigWhaleLabs/seal-cred-frontend/pull/452/files#diff-5b1516035d380825740282f22b8ac9801995a54f0ccbba08ab46bd23de3c8bdbR73)

src/components/badges/BadgeTitle.tsx
```diff
@@ -1,26 +1,24 @@
...
+import ExternalLink from 'components/ui/ExternalLink'
 import Network from 'models/Network'
-import getEtherscanAddressUrl from 'helpers/getEtherscanAddressUrl'
+import Proof from 'models/Proof'
+import getEtherscanAddressUrl from 'helpers/network/getEtherscanAddressUrl'
 
-function ProofName({ badge }: { badge: BaseProof }) {
-  if (badge instanceof ERC721Proof)
+function ProofName({ proof }: { proof: Proof }) {
+  if (isAddress(proof.original))
```
T-Damer, line 9: You're using `proof.original` two times, maybe it's better to destructure it on the top?

src/components/badges/BadgeTitle.tsx
```diff
@@ -38,5 +36,5 @@ export default function ({
 
   if (!proof) return null
 
-  return <ProofName badge={proof} />
+  return <ProofName proof={proof} />
```
T-Damer, line 39: Pass only `original` here

src/components/email-flow/EmailFlowListProofs.tsx
```diff
@@ -1,25 +1,30 @@
...
 import proofStore from 'stores/ProofStore'
 
 export default function ({
   onSelectProof,
 }: {
-  onSelectProof: (proof: EmailProof) => void
+  onSelectProof: (proof: ProofModel) => void
 }) {
-  const { emailProofsCompleted } = useSnapshot(proofStore)
+  const { proofsCompleted } = useSnapshot(proofStore['Email'])
```
T-Damer, line 13: Let's add a blank line before `return`

src/components/email-flow/EmailFlowProof.tsx
```diff
@@ -1,15 +1,15 @@
...
+import Proof from 'components/proofs/Proof'
+import ProofModel from 'models/Proof'
 
 export default function ({
   proof,
   onMinted,
   onMintFailed,
 }: {
-  proof: EmailProof
+  proof: ProofModel
```
T-Damer, line 12: Same interface as [here](https://github.com/BigWhaleLabs/seal-cred-frontend/pull/452/files#diff-5b1516035d380825740282f22b8ac9801995a54f0ccbba08ab46bd23de3c8bdbR73)

src/components/owned/OwnerInfo.tsx
```diff
@@ -29,82 +28,73 @@ const walletBox = classnames(
...
-                hyphens
-                address={badge.address}
-                network={Network.Goerli}
-              />
-            </AccentText>
-          </ExternalLink>
-        </HeaderText>
-      </>
+function replaceKeywordsInString(
+  template: string,
```
T-Damer, line 32: Can we make it consume an object arguments and extract it into `helpers`?

src/components/owned/OwnerInfo.tsx
```diff
@@ -29,82 +28,73 @@ const walletBox = classnames(
...
+  keyword: string,
+  element: JSXInternal.Element
+) {
+  return template
+    .split(keyword)
+    .reduce(
+      (chain, part, index) =>
+        index === 0 ? chain.concat(part) : chain.concat(element).concat(part),
```
T-Damer, line 40:
```TS
const concatedPart = chain.concat(part)
if (index) return concatedPart.concat(element)
return concatedPart
```
?

src/components/proofs/ERC721ProofSection.tsx
T-Damer: Yes, let's remove the name here

src/components/proofs/ERC721ProofSection.tsx
```diff
@@ -1,73 +1,55 @@
...
-  account: string
-  network: Network
+  dataKey: DataKey
 }) {
+  const { proofTitle } = data[dataKey]
   return (
-    <Suspense
-      fallback={
-        <ProofSection title={<BodyText>{network} NFTs</BodyText>}>
+    <ProofSection title={<BodyText>{proofTitle}</BodyText>}>
```
T-Damer, line 41: Lol, I did same thing in my PR )0))

src/components/proofs/EmailProofForm.tsx
```diff
@@ -60,10 +61,12 @@ export default function ({
...
-      if (secret) {
-        const proof = await ProofStore.generateEmail(domain, secret)
-        if (proof) onCreate(proof)
-      }
+      const proof = await data['Email'].createProof(
+        proofStore['Email'],
+        domain,
+        { secret }
+      )
+      if (proof) onCreate(proof)
```
T-Damer, line 69: Actually, if there's no proof, is this an error?

src/components/proofs/ProofsList.tsx
```diff
@@ -0,0 +1,60 @@
...
+  originals,
+  nothingToGenerateText,
+}: {
+  originals?: string[]
+  proofs: ProofModel[]
+  dataKey: DataKey
+  onCreate?: (original: string) => Promise<void>
+  nothingToGenerateText?: string
+}) {
+  if (proofs.length === 0 && originals?.length === 0) {
```
T-Damer, line 27:
```TS
if (proofs.length === 0 && originals?.length === 0 && nothingToGenerateText) return <HintCard>...</HintCard>
```
?

src/helpers/network/getContractName.ts
```diff
@@ -0,0 +1,31 @@
+import { useSnapshot } from 'valtio'
+import ContractMetadataStore from 'stores/ContractMetadataStore'
+import Network from 'models/Network'
+import networks from 'networks'
+import prettifyContractName from 'helpers/network/prettifyContractName'
+
+export default function (
+  address: string,
```
T-Damer, line 8: Why don't we put everything into an object? It will make easier for us to use this function

src/components/ui/ContractName.tsx
```diff
@@ -41,20 +38,10 @@ function ContractNameSuspended({
...
-  if (clearType) {
-    if (contractName && emailDerivativeContracts.includes(address))
-      contractName = contractName.replace(' email', '')
-
-    if (contractName && ERC721derivativeContracts.includes(address))
-      contractName = contractName.replace(' (derivative)', '')
-  }
-
-  contractName = prettifyContractName(contractName, truncate)
+  const contractName = getContractName(address, network, {
```
T-Damer, line 41: Let's put everything into an object argument

src/helpers/proofs/generateERC721.ts
```diff
@@ -0,0 +1,53 @@
...
+import walletStore from 'stores/WalletStore'
+
+export default async function generateERC721(
+  store: ProofStore,
+  original: string,
+  network: Network
+) {
+  try {
+    const account = walletStore.account
+    if (!account) {
```
T-Damer, line 21: Let's get rid of `{}`

src/helpers/proofs/generateERC721.ts
```diff
@@ -0,0 +1,53 @@
...
+import handleError from 'helpers/handleError'
+import walletStore from 'stores/WalletStore'
+
+export default async function generateERC721(
+  store: ProofStore,
+  original: string,
+  network: Network
+) {
+  try {
+    const account = walletStore.account
```
T-Damer, line 20: btw, do we need to use capital `WalletStore` here?

src/helpers/providers/createProvider.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export default function (
+  network: Network,
+  networkToRPC: { [network in 'heavy' | 'default']: string }
+) {
+  return {
+    network,
+    defaultProvider: new providers.JsonRpcProvider(
+      networkToRPC.default,
+      network.toLowerCase()
```
T-Damer, line 13: How about extracting `network.toLowerCase()` into `const`?

src/hooks/useBadge.ts
```diff
@@ -0,0 +1,28 @@
...
+  derivative: string
+  original: string
+}
+
+export default function (derivative: string) {
+  const { ledgers } = useSnapshot(SealCredStore)
+
+  return useMemo(() => {
+    for (const [ledgerName, ledger] of Object.entries(ledgers)) {
+      for (const original of Object.keys(ledger)) {
```
T-Damer, line 17: You can get rid of`{}` here

src/hooks/useContractsOwned.ts
```diff
@@ -1,10 +1,19 @@
...
+import Network from 'models/Network'
 
-export default function (store: ContractsStore) {
-  const { addressToTokenIds } = useSnapshot(store)
+export default function (networkName?: Network) {
+  const { networks } = useSnapshot(ContractsNetworkStore)
 
-  if (!addressToTokenIds) return []
+  const addressToTokenIds = (Object.keys(networks) as Network[])
+    .filter((network) => !networkName || networkName === network)
```
T-Damer, line 9: Not sure what is faster, doing `filter` + `reduce`, or doing:
```TS
.reduce((combined, netwrokName) => {
  if (networkName || !(networkName === networkName)) return
  return {
        ...combined,
        ...networks[networkName].addressToTokenIds,
      }
}
```
Something like this, I mean having `if` statement inside `reduce` 🤔

src/networks.ts
```diff
@@ -0,0 +1,15 @@
+import { ETH_RPC, ETH_RPC_MAINNET } from '@big-whale-labs/constants'
+import Network from 'models/Network'
+import createProvider from 'helpers/providers/createProvider'
+import env from 'helpers/env'
+
+export default {
+  [Network.Mainnet]: createProvider(Network.Mainnet, {
+    default: ETH_RPC_MAINNET,
```
T-Damer, line 8: Why do we use those from `constants` and not from `env.ts`? `env.ts` has those + you can override them if something goes wrong

src/networks.ts
```diff
@@ -0,0 +1,15 @@
...
+    heavy: env.VITE_ETH_RPC_MAINNET,
+  }),
+  [Network.Goerli]: createProvider(Network.Goerli, {
+    default: ETH_RPC,
```
T-Damer, line 12: Same question here

src/stores/ContractsStore.ts
```diff
@@ -1,48 +1,44 @@
...
-    mainnetDefaultProvider,
-    mainnetHeavyProvider,
-    Network.Mainnet
-  )
-).makePersistent(env.VITE_ENCRYPT_KEY)
+const ContractsNetworkStore = proxy({
+  networks: proxyNetworks,
+})
+
+export const BadgesNetwork = Network.Goerli
```
T-Damer, line 22: Do we need this? We know, that by default we're using `goerli` 🤔

src/stores/ContractsStore.ts
```diff
@@ -1,48 +1,44 @@
...
+export const BadgesContractsStore =
+  ContractsNetworkStore.networks[BadgesNetwork]
 
 subscribeKey(WalletStore, 'account', (account) => {
   if (account) {
-    void GoerliContractsStore.fetchMoreContractsOwned(account, true)
-    void MainnetContractsStore.fetchMoreContractsOwned(account, true)
+    for (const network of Object.values(ContractsNetworkStore.networks)) {
```
T-Damer, line 28: You can get rid of `{}` here

T-Damer (commented):
janitor-fist

src/stores/ContractsStore.ts
MixailE: And if something changes, we will definitely change everything right?

src/stores/ContractsStore.ts
MixailE: I'm not sure if it's a good idea to remove the {} in the for loop

src/networks.ts
MixailE: it's better to leave the constants for now

src/networks.ts
MixailE: it's better to leave the constants for now

src/hooks/useContractsOwned.ts
MixailE: here the processing speed is not important

src/hooks/useBadge.ts
MixailE: I'm not sure if it's a good idea to remove the {} in the for loop

src/helpers/proofs/generateERC721.ts
MixailE: We use lowercase variables and uppercase classes, I think this is true for stores too

src/components/owned/OwnerInfo.tsx
MixailE: Why would an object be better?

src/components/email-flow/EmailFlowProof.tsx
MixailE: I'm not sure if this should be put in a separate interface

src/components/badges/BadgeBlock.tsx
MixailE: We still need to think about this, perhaps this interface is not suitable for new proofs

src/helpers/network/prettifyContractName.ts
MixailE: Typically this function is used with addresses

src/components/owned/OwnerInfo.tsx
MixailE: Are they in the data file?

src/components/badges/List.tsx
```diff
@@ -67,33 +39,16 @@ function BadgeListSuspended() {
...
-            title="Email derivatives"
-            minted={ownedEmailDerivativeContracts}
-            proofs={proofsAvailableToMint.filter(
-              (proof) => proof instanceof EmailProof
-            )}
-            onMinted={onMinted}
-          />
+          {(Object.keys(data) as DataKey[]).map((ledgerName) => (
+            <BadgeSection
+              title={badgeConfig[data[ledgerName].badgeType].title(
```
T-Damer, line 44: Feels like we need to simplify this construction, maybe we can have a helper to extract the badge type from data?

T-Damer: The helper won't reduce the size of this, maybe simplification of the `data` and `badgeConfig` can help us? For example, can we extract `badgeConfig` using `ledgerName`?

src/components/email-flow/EmailFlowProof.tsx
```diff
@@ -25,7 +25,7 @@ export default function ({
     <>
       <HeaderText extraLeading>{headerTitle}</HeaderText>
       <BodyText>{statusText}</BodyText>
-      <ReadyEmailProof proof={proof} />
+      <Proof type="Email" original={proof.original} proof={proof} />
```
T-Damer, line 19: Hmm, why can't we extract `original` inside the component? You have a `proof` there

src/components/email-flow/EmailFlowListProofs.tsx
```diff
@@ -1,25 +1,31 @@
...
-      {Array.from(emailProofsCompleted).map((proof, index) => (
+      {Array.from(proofsCompleted).map((proof, index) => (
         <div
           onClick={() => onSelectProof(proof)}
           className={cursor('cursor-pointer')}
         >
-          <ReadyEmailProof proof={proof} key={`${proof.domain}-${index}`} />
+          <Proof
+            type="Email"
+            original={proof.original}
```
T-Damer, line 25: You have a proof, why do you need to pass `original` separately from it? Pass `proof` and get `original` inside the component

src/data.ts
```diff
@@ -0,0 +1,36 @@
...
+export enum BadgeSourceType {
+  Email = 'Email',
+  ERC721 = 'ERC721',
+}
+
+export default {
+  ExternalERC721: {
+    ledger: env.VITE_SC_EXTERNAL_ERC721_LEDGER_CONTRACT_ADDRESS,
+    badgeType: BadgeSourceType.ERC721,
```
T-Damer, line 17: How about to add here `badgeConfig: (badgeType) => badgeConfig[badgeType]`?

src/components/badges/List.tsx
```diff
@@ -67,33 +39,16 @@ function BadgeListSuspended() {
...
-          />
-          <BadgeSection
-            title="Email derivatives"
-            minted={ownedEmailDerivativeContracts}
-            proofs={proofsAvailableToMint.filter(
-              (proof) => proof instanceof EmailProof
-            )}
-            onMinted={onMinted}
-          />
+          {(Object.keys(data) as DataKey[]).map((ledgerName) => (
```
T-Damer, line 42: Looks like you're doing `Object.keys(data) as DataKey[]` a couple of times, maybe we can put it into `data.ts`:
```TS
export const dataKeys = Object.keys(data) as DataKey[]
```
?

T-Damer: You'll make some lines shorter

src/helpers/proofs/generateEmail.ts
```diff
@@ -0,0 +1,36 @@
...
+import { getEddsaPublicKey } from 'helpers/proofs/attestor'
+import { handleError } from '@big-whale-labs/frontend-utils'
+import buildEmailProof from 'helpers/proofs/buildEmailProof'
+import checkNavigator from 'helpers/proofs/checkNavigator'
+
+export default async function generateEmail(
+  store: ProofStore,
+  original: string,
+  options: {
+    secret?: string
```
T-Damer, line 13: Why don't we put `secret` outside of `options`?

src/helpers/proofs/generateERC721.ts
```diff
@@ -0,0 +1,55 @@
...
+import buildERC721Proof from 'helpers/proofs/buildERC721Proof'
+import checkNavigator from 'helpers/proofs/checkNavigator'
+import getNullifierMessage from 'helpers/proofs/getNullifierMessage'
+import walletStore from 'stores/WalletStore'
+
+export default async function generateERC721(
+  store: ProofStore,
+  original: string,
+  options: {
+    network: Network
```
T-Damer, line 19: Why don't we put `network` outside of `options`?

src/components/email-flow/EmailFlowProof.tsx
MixailE: Proof is optional parameter

src/stores/WalletStore.ts
```diff
@@ -159,40 +155,12 @@ class WalletStore extends PersistableStore {
     return signature
   }
 
-  async mintDerivative(proof: BaseProof) {
```
T-Damer, line 162: Where is the `mint` function now?

src/helpers/proofs/generateERC721.ts
MixailE: Because network and secret are not needed everywhere

src/helpers/proofs/generateEmail.ts
MixailE: Because network and secret are not needed everywhere

src/stores/WalletStore.ts
MixailE: I moved it to ProofStore

src/components/email-flow/EmailFlowListProofs.tsx
MixailE: This component is used in different places in different ways

T-Damer (approved):
lfg-lets

src/components/badges/List.tsx
MixailE: Okay, we'll have to think about it, let's fix it later

src/data.ts
MixailE: Will have to think about it in the future

PR #454, opened 2022-08-23
changed method of generating random signatures

tuansydau:
## What 
* Changed the way that we generate r2 and s2 (with `entropy-string` instead of `Crypto` library)
## Testing
* Created a few email, NFT, and GNFT proofs and badges
* Could not yet test on any machines that have previously had the `Failed to execute 'getRandomValues'` error


src/helpers/ERC721Proof.ts
```diff
@@ -108,7 +109,8 @@ export default class ERC721Proof
     balanceSignature: BalanceSignature,
     eddsaPublicKey: PublicKey
   ) {
-    const [r2, s2] = crypto.getRandomValues(new BigUint64Array(2))
```
T-Damer, line 111: If we don't use `crypto`, maybe we should remove it from `package.json`?

src/helpers/ERC721Proof.ts
```diff
@@ -108,7 +109,8 @@ export default class ERC721Proof
...
+    const r2 = new Entropy({ charset: '0123456789ABCDEF', bits: 64 }).string()
```
T-Damer, line 112: Hm, smells like it doesn't work: 
image

T-Damer: + you can extract `'0123456789ABCDEF'` into:
```TS
const charset = '0123456789ABCDEF'
```

tuansydau: Oh good idea

tuansydau:
Wondering if it's better to extract this into a helper function, since it's duplicate code? What do you think? @T-Damer

src/helpers/ERC721Proof.ts
```diff
@@ -108,7 +109,17 @@ export default class ERC721Proof
     balanceSignature: BalanceSignature,
     eddsaPublicKey: PublicKey
   ) {
-    const [r2, s2] = crypto.getRandomValues(new BigUint64Array(2))
+    const hexadecimalCharset = '0123456789ABCDEF'
```
T-Damer, line 112: Are you sure that this won't cause any `leaks` or `tips` for hackers? Maybe we need to put it into `.env`?

T-Damer: + we use this 2 times, it's better to put it into one place, not have two constants

tuansydau: It'll give a tip to hackers but unless there's an exploit with `entropy-string`, we should be safe since we're making random 64 bit ints. I agree that the constant should be extracted though

T-Damer (approved):

PR #469, opened 2022-09-05
Update email frow via email

smyzin:
- created a new hook that takes params from URL
- updated flow with taken params
- updated negative flow (account doesn't connected)

image
image


src/components/proofs/ProofList.tsx
```diff
@@ -29,8 +31,18 @@ const bottomWrapper = classnames(
 )
 
 export function ProofListSuspended() {
-  const [category, setCategory] = useState<CategoriesTitles>('NFTs')
+  const params = useUrlParams()
+
+  const [category, setCategory] = useState<CategoriesTitles>(
```
MixailE, line 36: Let's make a parameter where a specific section will be specified, without such checks

src/components/proofs/ProofList.tsx
```diff
@@ -29,8 +31,18 @@ const bottomWrapper = classnames(
...
+    params ? 'Email' : 'NFTs'
+  )
   const { hasAnyProof } = useProofStore()
+  if (params) {
```
MixailE, line 40: It seems to me that such logic should be in the rendering of sections, and not in the general file

smyzin: Did it

src/components/proofs/ProofList.tsx
```diff
@@ -28,8 +28,14 @@ const bottomWrapper = classnames(
...
 
-export function ProofListSuspended() {
-  const [category, setCategory] = useState<CategoriesTitles>('NFTs')
+export function ProofListSuspended({
+  selectedCategory,
+}: {
+  selectedCategory?: CategoriesTitles
+}) {
+  const [category, setCategory] = useState<CategoriesTitles>(
+    selectedCategory || 'NFTs'
```
MixailE, line 37: can we specify the first category from the object here?

src/components/proofs/ConnectAccount.tsx
```diff
@@ -49,6 +51,13 @@ const hintWrapper = classnames(
 
 export default function () {
   const { walletLoading, needNetworkChange } = useSnapshot(WalletStore)
+  const params = useUrlParams()
+  if (params) {
```
T-Damer, line 55: Let's remove `{}` from here

src/components/proofs/EmailProof.tsx
```diff
@@ -81,12 +81,29 @@ const questionBlock = (open: boolean) =>
...
+
+  useEffect(() => {
+    if (defaultToken && defaultDomain) {
+      EmailDomainStore.emailDomain = defaultDomain
+      setToken(defaultToken ?? '')
+      setError(undefined)
+      setDomain(defaultDomain)
+    }
+  }, [defaultDomain, defaultToken])
+  console.log(EmailDomainStore.emailDomain, defaultDomain)
```
T-Damer, line 106: I think we don't need this `log`

src/components/proofs/EmailProof.tsx
```diff
@@ -144,6 +162,7 @@ export default function () {
         {open && (
           <EmailProofForm
             domain={domain}
+            token={EmailDomainStore.emailDomain === defaultDomain ? token : ''}
```
T-Damer, line 165: Why don't we use `emailDomain` here? Without accessing `EmailDomainStore` I mean

src/components/proofs/EmailProof.tsx
```diff
@@ -81,12 +81,29 @@ const questionBlock = (open: boolean) =>
...
+  const { emailDomain } = useSnapshot(EmailDomainStore)
+  const [domain, setDomain] = useState(defaultDomain ?? '')
+  const [open, setOpen] = useState(!!defaultDomain)
   const [error, setError] = useState<string | undefined>()
   const [generationStarted, setGenerationStarted] = useState(false)
-  const { emailDomain } = useSnapshot(EmailDomainStore)
+  const [token, setToken] = useState(defaultToken ?? '')
+
+  useEffect(() => {
+    if (defaultToken && defaultDomain) {
```
T-Damer, line 99: `if (!defaultToken || !defaultDomain) return`
Less nesting = better code, usually, let's keep it

src/components/proofs/EmailProof.tsx
smyzin: Right, thx

src/components/proofs/EmailProof.tsx
```diff
@@ -81,12 +81,29 @@ const questionBlock = (open: boolean) =>
...
-  const [open, setOpen] = useState(false)
+export default function ({
+  domain: defaultDomain,
+  token: defaultToken,
+}: {
+  domain?: string
+  token?: string
+}) {
+  const { emailDomain } = useSnapshot(EmailDomainStore)
+  const [domain, setDomain] = useState(defaultDomain ?? '')
```
T-Damer, line 92: So, `defaultDomain` may be empty? Btw, you're doing this operation twice, worth extracting into `const`

src/components/proofs/EmailProof.tsx
smyzin: Changed it, thx

src/components/proofs/EmailProof.tsx
smyzin: Changed it too

src/components/proofs/ProofList.tsx
```diff
@@ -28,8 +28,12 @@ const bottomWrapper = classnames(
   alignItems('items-end')
 )
 
-export function ProofListSuspended() {
-  const [category, setCategory] = useState<CategoriesTitles>('NFTs')
+export function ProofListSuspended({
+  selectedCategory,
+}: {
+  selectedCategory: CategoriesTitles
```
T-Damer, line 34: I think it's better to extract this one into `interface`

smyzin: Done

src/components/proofs/ProofList.tsx
```diff
@@ -64,10 +68,14 @@ export function ProofListSuspended() {
   )
 }
 
-export default function () {
+export default function ({
+  selectedCategory,
+}: {
+  selectedCategory: CategoriesTitles
```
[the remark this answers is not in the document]
smyzin: Done

src/components/proofs/EmailProof.tsx
```diff
@@ -81,12 +81,28 @@ const questionBlock = (open: boolean) =>
...
```
T-Damer, line 92: You're checking `??` in `EmailProofSection` - the only place where we use `<EmailProof />`, I think that you need to:
1) Remove `?` from arguments
2) Remove checks from here, you're passing 100% string

T-Damer (approved):

PR #470, opened 2022-09-05
add warning about linked wallet

upacyxou:
- warning about linked wallet added

### DIFF
--- a/src/components/ui/LinkedWalletWarning.tsx
+++ b/src/components/ui/LinkedWalletWarning.tsx
@@ -1,0 +1,37 @@
+import { AccentText } from 'components/ui/Text'
+import Mark, { MarkType } from 'components/ui/Mark'
+import ToolTip from 'components/ui/ToolTip'
+import classnames, {
+  alignItems,
+  display,
+  gap,
+  margin,
+} from 'classnames/tailwind'
+
+const hintWrapper = classnames(
+  display('inline-flex'),
+  alignItems('items-center'),
+  gap('gap-x-1.5'),
+  margin('mt-3')
+)
+
+export default function () {
+  return (
+    <span>
+      <AccentText color="text-error">
+        <ToolTip
+          position="bottom"
+          fitContainer
+          text={
+            'The wallet that’s currently linked contains the original nft. It might be best you use a different wallet.'
+          }
+        >
+          <span className={hintWrapper}>
+            <Mark medium mark={MarkType.Exclamation} />
+            Wait
+          </span>
+        </ToolTip>
+      </AccentText>
+    </span>
+  )
+}
--- a/src/components/ui/Mark.tsx
+++ b/src/components/ui/Mark.tsx
@@ -23,26 +23,37 @@ const questionStyles = (small?: boolean, disabled?: boolean) =>
     cursor(disabled ? undefined : 'cursor-pointer')
   )
 
-const borderWrapper = (small?: boolean) =>
+const borderWrapper = (small?: boolean, medium?: boolean) =>
   classnames(
     boxSizing('box-content'),
     borderRadius('rounded-full'),
     borderWidth('border'),
     borderColor('border-current'),
-    width(small ? 'w-4' : 'w-6'),
-    height(small ? 'h-4' : 'h-6')
+    width({ 'w-4': small, 'w-5': medium, 'w-6': !small && !medium }),
+    height({ 'h-4': small, 'h-5': medium, 'h-6': !small && !medium }),
+    display('flex'),
+    justifyContent('justify-center')
   )
+
+export enum MarkType {
+  Question = '?',
+  Exclamation = '!',
+}
 
 export default function ({
   small,
   disabled,
+  medium,
+  mark = MarkType.Question,
 }: {
   small?: boolean
+  medium?: boolean
   disabled?: boolean
+  mark?: MarkType
 }) {
   return (
-    <div className={borderWrapper(small)}>
-      <div className={questionStyles(small, disabled)}>?</div>
+    <div className={borderWrapper(small, medium)}>
+      <div className={questionStyles(small, disabled)}>{mark}</div>
     </div>
   )
 }
--- a/src/components/badges/BadgeBlock.tsx
+++ b/src/components/badges/BadgeBlock.tsx
@@ -7,9 +7,12 @@ import BadgeCard from 'components/badges/BadgeCard'
 import BadgeTitle from 'components/badges/BadgeTitle'
 import BadgeWrapper from 'components/badges/BadgeWrapper'
 import Button from 'components/ui/Button'
+import LinkedWalletWarning from 'components/ui/LinkedWalletWarning'
 import ProofStore from 'stores/ProofStore'
 import WalletStore from 'stores/WalletStore'
 import badgeConfig from 'badgeConfig'
+import data from 'data'
+import useOwnedAddresses from 'hooks/useOwnedAddresses'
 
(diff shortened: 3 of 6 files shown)


src/components/ui/LinkedWalletWarning.tsx
```diff
@@ -0,0 +1,37 @@
...
+)
+
+export default function () {
+  return (
+    <span>
+      <AccentText color="text-error">
+        <ToolTip
+          position="bottom"
+          fitContainer
+          text={
```
T-Damer, line 25: Let's remove `{}` from here, eslint will come soon for this rule :)

src/components/ui/LinkedWalletWarning.tsx
```diff
@@ -0,0 +1,37 @@
...
+const hintWrapper = classnames(
+  display('inline-flex'),
+  alignItems('items-center'),
+  gap('gap-x-1.5'),
+  margin('mt-3')
+)
+
+export default function () {
+  return (
+    <span>
```
T-Damer, line 20: I think you don't need this `span`

src/components/ui/Mark.tsx
```diff
@@ -23,26 +23,37 @@ const questionStyles = (small?: boolean, disabled?: boolean) =>
     cursor(disabled ? undefined : 'cursor-pointer')
   )
 
-const borderWrapper = (small?: boolean) =>
+const borderWrapper = (small?: boolean, medium?: boolean) =>
```
T-Damer, line 26: This might grow later, some of us will add `big?:boolean, big-med, small-med etc.`, we need to think how to avoid it

src/components/ui/Mark.tsx
```diff
@@ -23,26 +23,37 @@ const questionStyles = (small?: boolean, disabled?: boolean) =>
...
   classnames(
     boxSizing('box-content'),
     borderRadius('rounded-full'),
     borderWidth('border'),
     borderColor('border-current'),
-    width(small ? 'w-4' : 'w-6'),
-    height(small ? 'h-4' : 'h-6')
+    width({ 'w-4': small, 'w-5': medium, 'w-6': !small && !medium }),
```
T-Damer, line 32: And I think this logic can be simplified, let's think how we can do this

src/components/ui/LinkedWalletWarning.tsx
upacyxou: ty

src/components/ui/LinkedWalletWarning.tsx
upacyxou: you're right, ty

src/components/ui/Mark.tsx
upacyxou: I tried to simplify it, have no idea how to simplify it more

src/components/ui/Mark.tsx
upacyxou: I tried to simplify it, don't have any workarounds in my mind

src/components/ui/Mark.tsx
```diff
@@ -0,0 +1,70 @@
...
+  justifyContent,
+  width,
+} from 'classnames/tailwind'
+
+const questionStyles = (size?: Sizes, disabled?: boolean) =>
+  classnames(
+    display('flex'),
+    justifyContent('justify-center'),
+    alignItems('items-center'),
+    fontSize(size === Sizes.Small ? 'text-xs' : 'text-base'),
```
T-Damer, line 21: `base` is default, use `{'text-xs': Sizes.Small}`

src/components/ui/Mark.tsx
```diff
@@ -0,0 +1,70 @@
...
+    fontFamily('font-primary'),
+    cursor(disabled ? undefined : 'cursor-pointer')
```
T-Damer, line 23: For one prop object is preferred: `{'cursor-pointer': !disabled}`

src/components/ui/Mark.tsx
```diff
@@ -0,0 +1,70 @@
...
+  )
+
+const borderWrapper = (size?: Sizes) => {
+  return classnames(
+    boxSizing('box-content'),
+    borderRadius('rounded-full'),
+    borderWidth('border'),
+    borderColor('border-current'),
+    width({
```
T-Damer, line 32: You have a square, this means you can remove `height` or `width` and set `h-auto`/`w-auto`

src/components/ui/Mark.tsx
```diff
@@ -0,0 +1,70 @@
...
+    display('flex'),
+    justifyContent('justify-center')
+  )
+}
+export enum MarkType {
+  Question = '?',
+  Exclamation = '!',
+}
+
+export enum Sizes {
```
T-Damer, line 51: Worth extracting into `models`

src/components/ui/Mark.tsx
```diff
@@ -0,0 +1,70 @@
...
+  Small = 'Small',
+  Medium = 'Medium',
+}
+
+export default function ({
```
T-Damer, line 56: Maybe we should have a better name for this component? Something like `CharInCircle`/`MarkInCircle`?

T-Damer: Because mark can be a square or triangle

smyzin (approved):
Only a small comment left

src/components/ui/Mark.tsx
```diff
@@ -0,0 +1,70 @@
...
+    display('flex'),
+    justifyContent('justify-center'),
+    alignItems('items-center'),
+    fontSize(size === Sizes.Small ? 'text-xs' : 'text-base'),
+    fontFamily('font-primary'),
+    cursor(disabled ? undefined : 'cursor-pointer')
+  )
+
+const borderWrapper = (size?: Sizes) => {
+  return classnames(
```
T-Damer, line 27: You don't need `return` here, simply do:
```TS
const borderWrapper = (size?: Sizes) => classnames(
```

src/components/ui/Mark.tsx
T-Damer: by square I mean that your box has w=h, not that the mark is square :)

T-Damer (approved):

src/components/ui/Mark.tsx
upacyxou: ty

src/components/ui/Mark.tsx
upacyxou: ty

src/components/ui/Mark.tsx
```diff
@@ -0,0 +1,70 @@
...
+}
+export enum MarkType {
+  Question = '?',
+  Exclamation = '!',
+}
+
+export enum Sizes {
+  Small = 'Small',
+  Medium = 'Medium',
+}
```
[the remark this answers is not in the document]
upacyxou: I think you're right, ty for catch

src/components/ui/Mark.tsx
upacyxou: ty

src/components/ui/Mark.tsx
upacyxou: ty, fixed

src/components/ui/Mark.tsx
upacyxou: this didn't work, unfortunately

src/components/ui/Mark.tsx
upacyxou: You're right, ty. I'll name it to `CharInCircle`

smyzin (approved):
Lgtm

T-Damer (approved):

PR #473, opened 2022-09-06
Add SealVerse dropdown that will help to open another project

T-Damer:


tailwind.config.js
```diff
@@ -57,9 +57,10 @@ module.exports = {
       },
       screens: {
         xxs: '17.6rem',
-        xs: '22.5rem',
+        xs: '22.6rem',
```
smyzin, line 60: Why do we need this change? Won't it break smt in the other place?

src/helpers/visibilityClassnames.ts
```diff
@@ -15,6 +15,8 @@ export const displayFrom = (from: Size) => {
       return display('hidden', 'sm:flex')
     case 'md':
       return display('hidden', 'md:flex')
+    case 'md700':
+      return display('hidden', 'md700:flex')
```
smyzin, line 19: What's mean `md700`? May we escape from using it?

src/components/navbar/SealVerse.tsx
```diff
@@ -0,0 +1,10 @@
+import Dropdown from 'components/Dropdown'
+
+export default function () {
+  const options = [
+    { label: 'SealCred', selected: true },
+    { label: 'SealCred Echo', href: 'https://echo.sealcred.xyz' },
+  ]
```
smyzin, line 7: Maybe we can put it into the config like `data`? Or include it in the existing file? Wdyt?

src/components/navbar/Navbar.tsx
```diff
@@ -41,13 +42,17 @@ const logoContainer = classnames(
...
+  displayFrom('md')
+)
 
-const logoWrapper = classnames(display('flex'), width('w-1/4', 'sm:w-full'))
+const logoWrapper = classnames(display('flex'), width('w-full'))
 
 export default function () {
   const { pathname } = useLocation()
-  const withoutWallet = pathname.split('/').length >= 3
+  const withoutRightBlock = pathname.split('/').length >= 3
```
smyzin, line 55: Maybe it's better to make it via `Layout`? But I guess it should be a separate task...

src/components/proofs/MobileProofCategories.tsx
```diff
@@ -21,6 +21,8 @@ export default function ({
       onChange={(selectedValue) => {
         setCategory(selectedValue as CategoriesTitles)
       }}
+      colorfulCurrentValue
+      hideAfterMd
```
smyzin, line 25: I guess we should do it without a prop. It just unnecessary

src/components/navbar/SealVerse.tsx
```diff
@@ -0,0 +1,10 @@
+import Dropdown from 'components/Dropdown'
+
+export default function () {
+  const options = [
+    { label: 'SealCred', selected: true },
```
smyzin, line 5: May we rid of the `selected`? Or checking it in another way, like getting the host and comparing it with the link from the option? It looks strange because we have a check if the option selected inside and also give the component `selected`

src/components/Dropdown/Option.ts
```diff
@@ -1,4 +1,6 @@
 export default interface Option {
   label: string
-  disabled: boolean | undefined
+  href?: string
+  selected?: boolean
+  disabled?: boolean
```
smyzin, line 5: I don't like the idea of setting `href` & `selected`. May we use `label` and `value` instead

src/components/Dropdown/Menu.tsx
```diff
@@ -38,29 +45,37 @@ const menuItem = (selected?: boolean) =>
...
             key={option.label}
-            className={menuItem(option.label === selected)}
+            className={menuItem(
+              option.selected || option.label === selected,
+              fitToItemSize
+            )}
             onClick={() => {
-              onSelect(option)
+              if (onSelect) onSelect(option)
+              if (option.href) openLinkInNewTab(option.href)
```
smyzin, line 78: We can open the link `onSelect`. I don't think we should check if an option has `href`

T-Damer: You're right, I only need to understand how to specify `onSelect` only for one option, not for each like now

src/components/Dropdown/Option.ts
T-Damer: Let's name it `forceSelected`

src/components/navbar/Navbar.tsx
T-Damer: Let's make a task for it

tailwind.config.js
T-Damer: We need it to optimize apply styles on 360px as well (not just 359px)
image

T-Damer: I double-checked, all pages work fine both on <360px and >360px

src/helpers/visibilityClassnames.ts
T-Damer: It helps us to remove the discord-twitter buttons on `width=700`, when right and left sides of navbar touch each other, not sure if we can remove it (maybe there's a way to detect intersection between elements and remove one-by-one, but we want to remove whole social-links block, probably)

src/components/navbar/SealVerse.tsx
T-Damer: Well, I think we can throw the `selected` logic inside the component

src/components/proofs/MobileProofCategories.tsx
T-Damer: We need it to keep things both on mobile and desktops, removing it means:
1) Make a default `hideAfterMd = true`
2) Add `hideAfterMd={false}` in other places

So let's keep it, better have a `true` prop, than `prop={false}`

src/sealVerseData.ts
```diff
@@ -0,0 +1,9 @@
+import openLinkInNewTab from 'helpers/openLinkInNewTab'
+
+export default [
+  { label: 'SealCred' },
+  {
+    label: 'SealCred Echo',
+    onSelect: () => openLinkInNewTab('https://echo.sealcred.xyz'),
```
MixailE, line 7: It would be easier/better to specify the link in the config

src/components/proofs/MobileProofCategories.tsx
smyzin: I mean that you need it only in one place, so you can wrap this dropdown into `div` with a class that will display on certain screens. You don't need to pass this prop to Dropdown.

upacyxou (approved):
lgtm

MixailE (approved):

src/components/Dropdown/index.tsx
```diff
@@ -80,10 +90,11 @@ export default function ({
         open={open}
         options={options}
         selected={currentValue}
-        onSelect={(option) => {
-          onChange(option.label)
+        onSelect={({ value, label }) => {
+          onChange(value || label)
```
smyzin, line 94: I guess we may left it as `onChange(value)`

smyzin (approved):
Let's prettify and merge it 🙂

src/components/navbar/RightBlock.tsx
```diff
@@ -85,6 +90,8 @@ export default function () {
         </SocialLink>
         <hr className={delimiterContainer} />
       </div>
+      <SealVerse />
+      <hr className={lastDelimiterContainer} />
```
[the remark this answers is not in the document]
T-Damer: image
Not, unfortunately, the box sizes are different

src/sealVerseData.ts
```diff
@@ -0,0 +1,9 @@
+import Option from 'components/Dropdown/Option'
+
+export default [
+  { label: 'SealCred', value: window.location.origin },
```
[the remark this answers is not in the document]
T-Damer: No, because this won't work on localhost, for example, if we'll change the link (sealc.red, for example), it won't work either

PR #474, opened 2022-09-07
refactor email-flow

MixailE:


src/components/email-flow/EmailFlowProof.tsx
```diff
@@ -1,9 +1,26 @@
...
+export default function ({
+  proof,
+  onReset,
+}: {
+  proof: ProofModel
+  onReset: () => void
+}) {
+  const { ledgers } = useSnapshot(SealCredStore)
+  const emailLedger = ledgers['Email']
+  const address = emailLedger[proof.original]
```
T-Damer, line 20: Let's specify which `address` this is, probably `derivativeAddress`

T-Damer (approved):

smyzin (approved):
@T-Damer is right, let's rename one const. Otherwise looks good to me

PR #515, opened 2023-04-13
Add cookie message

smyzin:
- added a new `Cookie` component
- added store value
- updated styles
- added a new `Text` component

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -32,7 +32,7 @@ const pageContainer = classnames(
 const bodyContainer = classnames(
   display('flex'),
   flexDirection('flex-col'),
-  margin('my-7', 'xxs:mx-auto', 'tablet:mt-0', 'tablet:mb-auto'),
+  margin('my-7', 'tablet:mt-0', 'tablet:mb-auto'),
   overflow('overflow-x-clip', 'sm:overflow-x-visible')
 )
 
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,6 +2,7 @@ import { Route, BrowserRouter as Router, Routes } from 'react-router-dom'
 import { ToastContainer } from 'react-toastify'
 import { lazy } from 'react'
 import Announcement from 'components/ui/Announcement'
+import Cookie from 'components/ui/Cookie'
 import Footer from 'components/ui/Footer'
 import LazyComponent from 'components/ui/LazyComponent'
 import Navbar from 'components/navbar/Navbar'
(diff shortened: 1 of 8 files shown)


smyzin:
[Add the use of a cookie message](https://trello.com/c/tD1YARKl/1809-add-the-use-of-a-cookie-message)

MixailE (approved):

src/components/ui/Cookie.tsx
```diff
@@ -0,0 +1,90 @@
...
+  zIndex,
+} from 'classnames/tailwind'
+
+const basicCardStyles = classnames(
+  display('flex'),
+  flexDirection('flex-col'),
+  position('fixed'),
+  inset(
+    'left-4',
+    'right-4',
```
T-Damer, line 30: You probably don't need both left and right

src/components/ui/Cookie.tsx
```diff
@@ -0,0 +1,90 @@
...
+  maxWidth('max-w-full', 'sm:max-w-cookie'),
+  zIndex('z-50')
+)
+const button = (small?: boolean) =>
+  classnames(
+    margin(small ? 'mx-auto' : 'mr-auto'),
+    small ? display('sm:hidden') : display('hidden', 'sm:block')
+  )
+
+const CookieButton = ({
```
T-Damer, line 56: Let's move it to a separate file

smyzin: Moved

src/components/ui/Cookie.tsx
smyzin: They are needed to make small gaps between the card and edges

MixailE (approved):

src/App.tsx
```diff
@@ -30,7 +32,8 @@ const pageContainer = classnames(
 const bodyContainer = classnames(
   display('flex'),
   flexDirection('flex-col'),
-  margin('my-7', 'xxs:mx-auto', 'tablet:mt-0', 'tablet:mb-auto')
+  margin('my-7', 'xxs:mx-auto', 'tablet:mt-0', 'tablet:mb-auto'),
+  overflow('overflow-x-clip', 'sm:overflow-x-visible')
```
T-Damer, line 36: Actually, we can clip it always, as I remember, nothing would break

smyzin: I checked it and didn't notice that smt is broken, but I remember that we had troubles with `overflow`, so I decided to return to default behavior to avoid any issues I didn't notice.

